PR #1, opened 2026-05-25
fix: use local timezone instead of UTC for displayed times

julian-prieto:
All "Updated HH:MM" timestamps were formatted from `DateTime<Utc>` without converting to the system's local timezone, causing a 3-hour offset (UTC-3) for users in that timezone.

Changes:
- Apply `.with_timezone(&chrono::Local)` before `.format()` in all 6 locations where times are displayed to users (widget tooltip, TUI panels, TUI status bar, and per-vendor renderers)
- Explicitly add the `clock` feature to the `chrono` dependency in `Cargo.toml`

Files changed:
- `Cargo.toml` — add `clock` feature to chrono
- `src/widget/render.rs` — tooltip updated time
- `src/tui/panels.rs` — panel updated time
- `src/tui/view.rs` — status bar refresh time
- `src/zai/vendor.rs` — tooltip updated time
- `src/openrouter/vendor.rs` — tooltip updated time
- `src/openai/vendor.rs` — tooltip updated time

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,7 +26,7 @@ serde_json = "1"
 toml = "0.8"
 toml_edit = "0.22"
 clap = { version = "4", features = ["derive", "env"] }
-chrono = { version = "0.4", features = ["serde"] }
+chrono = { version = "0.4", features = ["serde", "clock"] }
 fs2 = "0.4"
 tempfile = "3"
 directories = "5"
--- a/src/openai/vendor.rs
+++ b/src/openai/vendor.rs
@@ -244,7 +244,7 @@ fn render_tooltip(
     let updated = match outcome.cache_age {
         Some(age) => {
             let when = now - chrono::Duration::from_std(age).unwrap_or_default();
-            when.format("%H:%M").to_string()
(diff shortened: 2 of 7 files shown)


julian-prieto:
I didn't find any "Contribution" guide in the repository, please let me know if this is not the correct way to propose a change.

akitaonrails:
Thanks @julian-prieto for the PR. The local-time display fix was the right direction and is now merged and released in v0.4.1.

Before merging, I made a few follow-up fixes on top of your patch:

- Centralized the local timestamp formatting in `src/format.rs` instead of keeping separate `chrono::Local` calls in each renderer.
- Routed the widget, vendor tooltips, and TUI footer/panels through that shared helper.
- Fixed the snapshot tests so they still verify the actual local-time output while remaining deterministic across machines in different timezones.
- Ran the full gate: `cargo test`, `cargo fmt --check`, `cargo clippy --all-targets -- -D warnings`, `cargo machete`, and `cargo build --release`.

Fix/merge commit: https://github.com/akitaonrails/ai-usagebar/commit/a01f602544881754b281d8d44b1d121554f9323f

PR #8, opened 2026-06-04
feat: native Windows support

EaeDave:
## Summary

Makes the `ai-usagebar` and `ai-usagebar-tui` binaries build and run **natively on Windows** (`x86_64-pc-windows-msvc`), with no WSL. The changes are small and platform-portable; Linux/macOS behavior is unchanged.

The Waybar **widget** is Wayland-only and intentionally out of scope — on Windows the TUI (and `ai-usagebar --json` for custom bars) is the entry point.

## What changed

- **Cross-platform home resolution.** Credential paths now resolve the home directory through `directories::BaseDirs` via a new shared `cache::home_dir()` helper, instead of reading `$HOME` directly (unset on native Windows). Both OAuth vendors share the resolver the same way they already share `cache::atomic_write`:
  - Anthropic → `%USERPROFILE%\.claude\.credentials.json`
  - OpenAI Codex → `%USERPROFILE%\.codex\auth.json`
- **Waybar refresh gated to Unix.** `request_refresh()` shells out to `pkill -RTMIN+13 waybar`; off Unix there's no Waybar process and no `pkill`, so the body is `#[cfg(unix)]`-gated and the call becomes a no-op (consumers poll on their own interval).
- **TUI key handling on Windows Terminal.** Terminals that report key `Repeat`/`Release` in addition to `Press` (Windows Terminal; Kitty-keyboard emulators) made one Tab/arrow move several tabs and holding a key fly through them. The TUI now acts only on `KeyEventKind::Press`. Not cfg-gated — it's harmless and beneficial everywhere.
- **Tests.** `tests/live.rs` now resolves creds via the shared `default_path()` (it read `$HOME` directly, which would panic on Windows). The `write_back`/`save_to_path` unit tests pointed their target at a live `NamedTempFile` handle; Windows refuses to rename over an open file, so they now use a closed file inside a `TempDir` (the same pattern `cache.rs` already uses). Added `default_path()` tests, including `#[cfg(windows)]` assertions that `%USERPROFILE%` is honored. Production code is unchanged.

## Notes for review

- **No new dependencies.** `directories` (already a dep) is the only crate involved; TLS is already `rustls` (no OpenSSL), file locking is `fs2` (winapi-backed) — both already Windows-capable.
- **`cache::home_dir()` is the de-dup choice**: it lives next to `xdg_cache_dir()` and is reused by both creds vendors so home resolution stays in one place.

## Testing

- Linux: `cargo test` (232), `cargo clippy --all-targets -- -D warnings`, `cargo fmt --check`, `cargo machete` — all clean.
- **Windows 11 (native, MSVC), Windows Terminal:** `cargo test` → **232 passed; 0 failed** (the extra test vs Linux is the Windows-gated `%USERPROFILE%` assertion). Verified `ai-usagebar.exe --pretty` reads real Anthropic usage and `ai-usagebar-tui.exe` navigates one tab per keypress.

## Out of scope

A separate Windows tray/UI consuming `ai-usagebar --json` lives outside this repo; this PR is backend-only.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,32 @@
+name: CI
+
+# Native Windows test signal for PRs. The Waybar widget is Wayland-only, so the
+# Linux gate lives in the maintainer's existing flow; this job proves the CLI +
+# TUI build and test on a real windows-latest runner (no WSL).
+on:
+  push:
+    branches: [main, windows-backend]
+  pull_request:
+  workflow_dispatch:
+
+concurrency:
+  group: ci-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  windows:
+    name: test (windows-latest)
+    runs-on: windows-latest
+    steps:
+      - uses: actions/checkout@v5
+
+      - uses: dtolnay/rust-toolchain@stable
+
+      - uses: Swatinem/rust-cache@v2
+
+      # ring assembles its primitives on x86_64 and needs nasm; the runner does
+      # not ship it on PATH. No-op if a future image already provides it.
+      - uses: ilammy/setup-nasm@v1
+
+      - name: Test
+        run: cargo test --all-targets --locked
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,26 @@ Each release is also published at
 
 ## [Unreleased]
 
-Nothing yet.
+### Added
+
+- **Native Windows support for the `ai-usagebar` and `ai-usagebar-tui`
+  binaries.** Credential paths now resolve the home directory through
+  `directories::BaseDirs` (a new shared `cache::home_dir()` helper) instead
+  of reading `$HOME` directly, so Anthropic (`%USERPROFILE%\.claude\.credentials.json`)
+  and OpenAI Codex (`%USERPROFILE%\.codex\auth.json`) credentials are found
+  natively on Windows. The Waybar refresh (`pkill -RTMIN+13 waybar`) is gated
+  to Unix and becomes a no-op elsewhere, since Waybar is Wayland-only. Linux
+  and macOS behavior is unchanged. The Waybar widget itself remains
+  Wayland-only; on Windows the TUI is the entry point.
+
+### Fixed
+
+- **TUI double-processed keystrokes on Windows Terminal.** Terminals that
+  report key `Repeat`/`Release` events in addition to `Press` (Windows
+  Terminal, and emulators advertising the Kitty keyboard protocol) made one
+  Tab/arrow press move several tabs and holding a key fly through them. The
+  TUI now acts only on `KeyEventKind::Press`. Harmless and beneficial on all
+  platforms.
 
 ## [0.5.1] — 2026-06-01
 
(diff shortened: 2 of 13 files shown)


akitaonrails:
Thanks for the Windows support work. I reviewed the diff and the Linux gate is clean (`cargo fmt --check`, `cargo test`, `cargo clippy --all-targets -- -D warnings`, `cargo machete`), but I would not merge this yet.

What is missing:

- A real Windows CI signal. This PR main claim is native Windows support, but the repository currently has no `windows-latest` PR check, and my local Windows target check could not complete on Linux because MSVC tooling (`lib.exe`) is unavailable. Please add or run a Windows job that at least does `cargo test --all-targets` or `cargo check --all-targets`.
- User-facing config messages still hard-code the Unix path and permission model, for example `~/.config/ai-usagebar/config.toml (chmod 600)`. On Windows the actual path comes from `directories::ProjectDirs`, so those messages should use the resolved path or be platform-neutral.
- The Windows-only `%USERPROFILE%` assertions are a little brittle because `BaseDirs` may return a known-folder/profile path with different casing or redirection. Prefer testing stable trailing path segments, or compare normalized/canonicalized paths if a prefix assertion is needed.

Once those are addressed, the code changes look small and directionally safe.

EaeDave:
Thanks for the thorough review — I've addressed all three points on `windows-backend`.

**1. Windows CI signal.** Added `.github/workflows/ci.yml` with a `windows-latest` job running `cargo test --all-targets --locked` (plus `ilammy/setup-nasm@v1`, since `ring` assembles its primitives on x86_64). It's green on my fork: https://github.com/EaeDave/ai-usagebar/actions/runs/27065094480 — **233 passed; 0 failed** on `windows-latest`. The workflow also triggers on `pull_request`, so it should run here once you approve workflows for the PR.

**2. Platform-neutral config messages.** Added `config::config_path_hint()`, which resolves the path through `directories::ProjectDirs`, and used it in the API-key error, the TUI "no vendors enabled" notice, the widget "vendor disabled" error, and the Settings "saved" status — no more hard-coded `~/.config/...`. The `(chmod 600)` note is now `#[cfg(unix)]`-only, since Windows has no such step.

**3. Brittle `%USERPROFILE%` assertion.** Replaced the `Path::starts_with` prefix check (component-wise case-sensitive, which can diverge from what `BaseDirs` returns via `SHGetKnownFolderPath`) with a normalized (lowercased, separator-normalized) comparison. The cross-platform `ends_with(".claude/.credentials.json")` test stays as the stable trailing-segment check. Both pass on the run linked above.

Gate stays clean: `cargo fmt --check`, `cargo clippy --all-targets -- -D warnings`, and `cargo test` (232 on Linux, 233 on Windows).

akitaonrails:
Thanks @EaeDave — clean, idiomatic cross-platform Rust. Standout calls: noticing the `$HOME` reads were about to be duplicated and consolidating into `cache::home_dir`, leaving the Press-only TUI fix ungated (correct: Linux terminals only emit Press), and the closed-file pattern for atomic-rename tests on Windows. Squash-merged as 7360307 (rolled into v0.6.0). Maintainer adjustments at merge time: kept PR #9's hermeticity invariant intact and added a one-line carve-out covering the Windows-gated `USERPROFILE` assertion (resolvers testing their own OS convention); dropped the dead `windows-backend` branch ref from the CI trigger.

Issue #2, opened 2026-05-26
Feature Proposal: Session Management


z4nder:
# Feature Proposal: Session Management / Resume Hub

## Problem

Tools like Codex and Claude support resuming sessions:

```bash
codex resume 019e6456-0b4d-7361-9f5c-xxx
```

```bash
claude --resume f66fece2-340e-4e60-b4f3-xxx
```

But these sessions are usually lost in terminal history and are hard to track later.

Users currently do not have an easy way to see:

- Existing sessions
- Token usage per session
- Last activity
- Related project/repository
- Quick resume commands
- Session history

---

## Proposal

Add a new `Sessions` view/tab to manage AI sessions.

Example:

```text
Provider   Project        Title                    Tokens      Last Used

Claude     teste-1   SRS improvements         128k        2h ago
Codex      teste-2     GitHub parser            345k        1d ago
Claude     teste-3       UI refactor              52k         3d ago
```

## Requirement / Feasibility Check

Before implementing this feature, it would be important to evaluate:

- If Codex, Claude and other tools expose session metadata locally
- If session IDs, token usage and timestamps can be extracted reliably
- If title/description information already exists
- If logs or local storage are stable enough for parsing
- If a common session model can be created across providers

If the data is not available directly, the project could still keep its own local session registry and enrich it over time.

z4nder:
I'm using the project on my NixOS + Hyprland setup and it has been really great so far.

I was already working on a project with a similar purpose, starting with cost tracking and with session management as a final goal.

I'll investigate the requirements mentioned above and see if this is feasible.

akitaonrails:
Thanks @z4nder for writing this up, and glad it has been useful on NixOS + Hyprland.

I did a first local feasibility pass without reading transcript contents.

What looks feasible:

- Codex looks like the best starting point. `~/.codex/state_5.sqlite` has a `threads` table with `id`, `rollout_path`, `created_at`, `updated_at`, `model_provider`, `cwd`, `title`, `tokens_used`, `git_sha`, `git_branch`, `git_origin_url`, `model`, and `preview`. That is enough for a read-only session list plus a `codex resume <id>` command.
[...]

PR #9, opened 2026-06-06
test: make vendor-resolution + TUI/theme tests hermetic (fix AUR source-build failure)

sombraSoft:
## The symptom: `yay -S ai-usagebar` fails to install — but only on *your* machine

A user who already runs the widget tries to update it from the AUR **source**
package:

```console
$ yay -S ai-usagebar
...
   Compiling ai-usagebar v0.5.1 (.../ai-usagebar-0.5.1)
    Finished `release` profile [optimized] target(s) in 30.15s
     Running unittests src/lib.rs (target/release/deps/ai_usagebar-1ecc9421f0b03946)
...
test widget::cli::tests::primary_from_config_wins_when_vendor_unset ... FAILED
test widget::cli::tests::defaults_match_claudebar ... FAILED

failures:

---- widget::cli::tests::primary_from_config_wins_when_vendor_unset stdout ----
thread '...' panicked at src/widget/cli.rs:212:9:
assertion `left == right` failed
  left: Openai
 right: Openrouter

---- widget::cli::tests::defaults_match_claudebar stdout ----
thread '...' panicked at src/widget/cli.rs:195:9:
assertion `left == right` failed
  left: Openai
 right: Anthropic

failures:
    widget::cli::tests::defaults_match_claudebar
    widget::cli::tests::primary_from_config_wins_when_vendor_unset

test result: FAILED. 227 passed; 2 failed; 0 ignored; 0 measured; 0 filtered out

error: test failed, to rerun pass `--lib`
==> ERROR: A failure occurred in check().
    Aborting...
```

<sub>*(Real output, reproduced at `v0.5.1` with `active_vendor=openai` — this is exactly what `makepkg` runs in `check()`. Build path elided.)*</sub>

The build aborts. On a teammate's fresh machine the same version installs fine.
The only difference: the affected user had **scroll-cycled the widget to a
non-default vendor** at some point, so `~/.cache/ai-usagebar/active_vendor`
holds `openai`.

This only bites **source-install** users. `packaging/aur/PKGBUILD` runs the test
suite in its `check()`:

```bash
check() {
  cd "$pkgname-$pkgver"
  cargo test --frozen --release
}
```

`makepkg` runs `check()` against the *building user's real `$HOME`*, so a
customized `active_vendor` file leaks into the tests. The `-bin` package skips
compilation and tests entirely, so `ai-usagebar-bin` users never hit it — which
is exactly why the failure looks non-reproducible.

## Root cause: tests reached real `$HOME`/`$XDG` state instead of an injected seam

Two unit tests asserted a *fixed* resolved vendor:

- `defaults_match_claudebar` → expects `Anthropic`
- `primary_from_config_wins_when_vendor_unset` → expects `Openrouter`

Both went through `Cli::resolved_vendor`, which calls `crate::active::read()` →
reads `~/.cache/ai-usagebar/active_vendor`. The widget's documented vendor
precedence is:

1. explicit `--vendor` (highest)
2. **persisted scroll-cycle state** (`active_vendor`)
3. `[ui] primary` from config
4. anthropic (lowest)

So a planted `active_vendor` at precedence level 2 overrides the value the
tests expect at levels 3–4 — and the assertion flips to the cycled vendor.

A short audit also found `App::new` reads the Omarchy theme file +
`$HOME` via `merged_with_omarchy()`. Its assertions don't depend on the palette
so it wasn't *failing*, but it's the same non-hermetic class of bug and would be
a latent flake.

### Reproduce on any clean checkout

```bash
mkdir -p ~/.cache/ai-usagebar && printf openai > ~/.cache/ai-usagebar/active_vendor
cargo test --lib widget::cli::tests   # 2 failures: left: Openai
rm -rf ~/.cache/ai-usagebar           # clean up
```

## The fix: inject the path/dependency, keep the real-path resolver as a thin wrapper

This repo already has the right pattern — `Cache::at(path)` for tests vs
`Cache::for_vendor()` for production. This PR applies the same seam to the three
offenders:

| Production (unchanged) | New test seam |
| --- | --- |
| `active::{read,write,cycle}` | `active::{read_from,write_to,cycle_at}` — take an explicit path |
| `Cli::resolved_vendor(config)` | `Cli::resolve_vendor_with(config, active)` — pure precedence, no disk |
| `App::new(vendors)` | `App::with_theme(vendors, theme)` — explicit palette, no theme-file read |

The real-path functions become one-line wrappers over the seam, so **production
behavior is unchanged**. The hermetic tests then exercise the logic
*deterministically*:

- scroll-cycle precedence is now covered on purpose (it was only ever exercised
  by accident, via whatever happened to be in the developer's real file);
- a `TempDir` round-trip covers `cycle_at`/`read_from`/`write_to`;
- the TUI/theme tests pass `Theme::default()` and never touch
  `~/.config/omarchy/...`.

`resolved_vendor` still short-circuits the `active_vendor` read when `--vendor`
is explicit (the documented multi-module Waybar config), so the disk-access
profile is identical to before — not just the output.

A **`Tests are hermetic`** invariant is added to `CLAUDE.md` so this class of
regression is caught in review rather than at install time.

## Verification

- `cargo test` → **235 lib + 3 e2e pass**, 4 live tests `#[ignore]`d; clean.
- `cargo clippy --all-targets -- -D warnings` → clean.
- Suite re-run with the **exact hostile state** that broke the install
  (`active_vendor=openai` planted in the real cache path) → **235/235 pass**.

## Scope

No `Cargo.toml` / `CHANGELOG.md` / PKGBUILD changes and no version bump — this is
a test-only + internal-seam change with no user-visible behavior difference.
Cutting a release is the maintainer's call (per the `CLAUDE.md` checklist); happy
to add a `CHANGELOG.md` "Fixed" entry if you'd like one.

---

**Diffstat:** 4 files, +197 / −35 (net +6 tests, 229 → 235).

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -81,6 +81,17 @@ patch version instead.
   the user's choice (and `chmod 600`ed by the Settings overlay), but
   **never commit** a real key. The `.gitignore` covers `.env`,
   `*.credentials.json`, and `.claude/`.
+- **Tests are hermetic.** A `#[test]`/`#[tokio::test]` must never read or
+  write a real `$HOME`/`$XDG` path (config, cache, creds, Omarchy theme)
+  or branch on an ambient env var — the AUR `check()` runs `cargo test`
+  during `makepkg`, so any test that reads a user's *customized* files
+  fails the install on their machine. Always inject the path/dependency
+  via the test seam, never the real-path resolver:
+  `Cache::at` not `for_vendor`, `active::{read_from,write_to,cycle_at}`
+  not `{read,write,cycle}`, `creds::read_from` not `default_path`,
+  `Theme::merged_with_omarchy_file` not `merged_with_omarchy`,
+  `Cli::resolve_vendor_with` not `resolved_vendor`, `App::with_theme`
+  not `new`. Live API tests stay behind `#[ignore]` (see `tests/live.rs`).
 
 ## Secret-discipline rules (learned the hard way)
 
--- a/src/active.rs
+++ b/src/active.rs
@@ -6,7 +6,7 @@
 //! at `<cache-dir>/active_vendor`.
 
 use std::fs;
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
 
 use crate::cache::atomic_write;
 use crate::error::{AppError, Result};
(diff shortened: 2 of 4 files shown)


akitaonrails:
Thanks @sombraSoft — exceptional bug-hunting work. The reproduction was airtight: tracing the AUR failure to precedence rule #2 + a planted `active_vendor` file, then proposing the seam pattern that already existed elsewhere in the codebase, made this trivial to land. Squash-merged as 7360307 (rolled into v0.6.0). A short follow-up commit on main pinned the user-facing CHANGELOG note (`Fixed: AUR source install no longer fails for users with a customized active_vendor`).

PR #10, opened 2026-06-12
fix(tui): signal Waybar after every live API fetch, not just Settings save

pedrohensto:
## Symptom

With the sample Waybar module (`interval: 300`, `signal: 13`), the bar can sit on a stale percentage — e.g. showing `cld 4%` — while the TUI launched from `on-click` shows the real, much higher number right next to it. The bar only catches up on its next 300-second tick, so the two UIs visibly disagree for minutes at a time.

## Root cause

The TUI already refreshes the **shared** on-disk cache (`~/.cache/ai-usagebar/<vendor>/usage.json`) on open, on `r`/`R`, and on its 60-second auto-refresh — but only the Settings overlay fired `SIGRTMIN+13` (`src/tui/settings.rs`). So fresh data lands on disk and Waybar is never told about it.

## Fix

- `ReadyTab` gains a `fresh_fetch` flag, true exactly when `fetch_snapshot` reported `cache_age == Some(Duration::ZERO)` — the existing convention all five vendors use to mean "these bytes came from the API during this call" (cache reuse and stale fallbacks report the payload's real mtime-derived age).
- The TUI binary calls `waybar::request_refresh()` when such a result lands, mirroring what the Settings overlay already does after a save.

Properties preserved:

- **No extra API calls.** The signaled re-exec hits the just-written cache well inside its 60s TTL, so the README's "keep Waybar polling at 300s" rate-limit guidance is unaffected. Cache-reuse and stale-fallback results don't signal at all.
- **No-op without the signal.** Modules not configured with `signal: 13` behave exactly as before (documented in the README addition).
- **Hermetic tests.** The decision is a pure helper (`is_live_fetch`) with a unit test; the `pkill` side effect lives only in the TUI binary's event loop, matching the existing `settings.rs` placement.

## Testing

- `cargo test` — 242 passed (includes the new `is_live_fetch_only_on_zero_cache_age`)
- Manually verified on Omarchy/Arch (waybar running, anthropic + openai enabled) with a logging shim in front of `pkill` on `PATH`:
  - caches older than the 60s TTL → opening the patched TUI logged `pkill -RTMIN+13 waybar` once per live-fetched vendor, and the running bar refreshed to the new percentage;
  - re-opening immediately (caches now fresh) → zero signals, confirming cache reuse stays silent.

CHANGELOG updated under `[Unreleased]`; README's signal section documents the new behavior.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,16 @@ Each release is also published at
 
 ## [Unreleased]
 
-Nothing yet.
+### Fixed
+
+- The TUI now fires `SIGRTMIN+13` after every refresh that hits the live API
+  (initial load, `r`/`R`, and the 60-second auto-refresh), so a Waybar module
+  configured with `signal: 13` re-execs against the freshly written cache and
+  shows the same numbers as the TUI. Previously only the Settings overlay
+  signaled Waybar, so the bar could keep displaying usage up to `interval`
+  seconds (300s in the sample config) older than the TUI open right next to
+  it. The signaled re-exec reuses the fresh cache, so no extra API call is
+  made.
 
 ## [0.7.1] — 2026-06-08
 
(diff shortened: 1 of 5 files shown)


pedrohensto:
Sorry bout that, I was investigating it on my local repo and Claude just opened a PR, incredible.
there's already a `exec-on-event` on waybar and it updates correctly.

Issue #14, opened 2026-07-03
Tracking multiple accounts for the same vendor?


zanlucathiago:
Hey — first off, thanks for this. I came over from claudebar and the extra vendors plus the TUI are great.

Quick feature request: I have a few Claude subscriptions and I'd like to watch usage for all of them at the same time. The config looks like it assumes one account per vendor (`[anthropic] credentials_path`), so I couldn't find a first-class way to track more than one.

Poking around the code I noticed `--creds-path` + `--cache-dir` already let me pull it off: point each at its own credentials file and run the binary once per account, each with its own cache dir so they don't clobber one another. Works really well. But both flags are `hide = true` / "for tests / debugging", so it feels like I'm leaning on a side door you didn't mean to expose.

Would you be open to making multiple accounts per vendor a supported thing? Rough shape:

```toml
[[anthropic.accounts]]
label = "personal"
credentials_path = "~/.config/ai-usagebar/accounts/personal.json"

[[anthropic.accounts]]
label = "work"
credentials_path = "~/.config/ai-usagebar/accounts/work.json"
```

with the label showing up in the bar text and one TUI tab per account. Beyond the config, the main change is keying the cache on account instead of just vendor — which the hidden `--cache-dir` already does under the hood.

Happy to send a PR if you're up for it. Just wanted to sort out the design and naming with you before writing anything, since it touches config and the TUI layout. Cheers.

akitaonrails:
Great report — everything you found is accurate, including the two hidden flags. One correction that happens to work in your favor: `--creds-path` currently applies to the **Anthropic vendor only**, which is exactly your use case.

**Done as of 39d9554** (will ship in the next release): `--cache-dir` and `--creds-path` are now documented, supported flags — visible in `--help`, pinned by a stability test, with a "Multiple accounts (advanced)" section in the README showing the exact two-module Waybar setup you're already running. Your side door is now a front door, and token refreshes write back to whichever file `--creds-path` names, so each account keeps itself alive independently.

**On first-class `[[anthropic.accounts]]`:** open to it, but it's more surface than it looks. Beyond config + cache keying it touches:

- the scroll-cycle state format (`active_vendor` persists a bare vendor slug today — account-qualified slugs need a back-compat parse),
- `--vendor anthropic` becoming ambiguous on the CLI (needs `--account <label>`),
- the Settings overlay (a flat per-vendor form today; editable account arrays are a different UI),
[...]

Issue #15, opened 2026-07-04
macOS: Anthropic refresh should fall back to Keychain when ~/.claude/.credentials.json is unusable


igorsdm:
## Summary

On macOS, `ai-usagebar` prefers `~/.claude/.credentials.json` whenever the file exists.

If that file contains stale or unusable Claude OAuth data, but newer valid credentials exist in the macOS Keychain (`Claude Code-credentials`), `ai-usagebar` still ignores Keychain. This causes Anthropic refresh to fail and stale cached usage to be reused.

## Current Behavior

- `~/.claude/.credentials.json` is used if it exists.
- Keychain is only used when the file is missing.
- If the file exists but contains unusable credentials, Anthropic refresh fails.
- Cached Anthropic usage may then be shown as `stale cache` in `ai-usagebar-tui`.

## Expected Behavior

On macOS, if `~/.claude/.credentials.json` exists but is clearly unusable, `ai-usagebar` should fall back to the `Claude Code-credentials` Keychain item instead of treating the file as authoritative.

## Steps To Reproduce

Prerequisites:

- `~/.claude/.credentials.json` exists.
- File credentials are stale or empty.
- macOS Keychain contains newer valid Claude Code credentials.

1. Run:

   ```bash
   ai-usagebar --vendor anthropic --json
   ```

2. Observe stale-cache or failed-refresh behavior.

3. Compare the default credentials path with a missing credentials path.

   Default path:

   ```bash
   ai-usagebar --vendor anthropic --json --cache-dir "$(mktemp -d)"
   ```

   Missing credentials path:

   ```bash
   ai-usagebar --vendor anthropic --json --cache-dir "$(mktemp -d)" --creds-path /path/that/does/not/exist
   ```

On my machine:

- The default path failed because the stale file was used.
- Forcing a missing credentials path succeeded because the code fell back to Keychain.

## Root Cause

Credential precedence on macOS is too strict:

- File existence is treated as sufficient reason to trust `~/.claude/.credentials.json`.
- Keychain fallback only happens when the file is missing.
- Unusable file contents do not trigger Keychain fallback.

## Proposed Change

In `src/anthropic/creds.rs`, keep file-first behavior by default, but add a macOS fallback path:

1. Read and parse `~/.claude/.credentials.json`.
2. If file credentials are usable, keep them.
3. If file credentials are clearly unusable, try `Claude Code-credentials` from Keychain.
4. If Keychain credentials are usable, prefer them.
[...]

PR #16, opened 2026-07-04
feat(anthropic): config-driven multiple accounts via --account (#14)

zanlucathiago:
Draft for #14 — first pass at first-class multiple accounts, scoped to the config + widget/CLI layer as we discussed. TUI tabs left for a follow-up.

### What's here

- `[[anthropic.accounts]]` in config (`label` + `credentials_path` — your sketch as-is).
- `--account <label>` on the CLI to pick one. Anthropic-only, and it conflicts with `--creds-path` since both name a credentials file.
- Named accounts get an isolated cache at `~/.cache/ai-usagebar/anthropic/<label>/` automatically — no `--cache-dir` needed.

### Sticking to the four back-compat rules

1. Singular `[anthropic] credentials_path` stays the default account, valid forever.
2. 0–1 accounts keep `~/.cache/ai-usagebar/anthropic/` — only *extra* accounts get a `<label>` subdir, no migration.
3. `--vendor anthropic` with no `--account` resolves to the default → byte-identical output.
4. TUI multi-tab is a follow-up; the widget + config layer here is complete.

### Tests / docs

Cover config parse + lookup (including the typo error that lists known labels), target resolution for default vs named vs unknown account, and the `--account`/`--creds-path` conflict. `cargo test` green (247), `clippy -D warnings` clean, `fmt` clean. README's "Multiple accounts" section gets a config-driven subsection; CHANGELOG under `[Unreleased]`.

Left as **draft** so you can sanity-check the config/CLI shape before I touch the TUI.

---

One unrelated heads-up while I was here: `cargo clippy` on macOS trips `collapsible_if` in `anthropic/creds.rs::write_back` (the Keychain-gated branch). CI doesn't catch it because that block is `#[cfg(target_os = "macos")]` and CI runs on Linux. Not touching it in this PR — happy to send a one-liner separately if you want it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,19 @@ Each release is also published at
 
 ## [Unreleased]
 
-Nothing yet.
+### Added
+
+- **Config-driven multiple Anthropic accounts** (#14). Declare extra
+  subscriptions once under `[[anthropic.accounts]]` (`label` +
+  `credentials_path`) and select one on the CLI with `--account <label>`,
+  instead of repeating `--creds-path`/`--cache-dir` on every widget module.
+  Each named account gets an isolated cache at
+  `~/.cache/ai-usagebar/anthropic/<label>/`. Anthropic-only; `--account`
+  conflicts with `--creds-path`. Fully back-compatible: the singular
+  `[anthropic] credentials_path` stays the default account, `--vendor
+  anthropic` with no `--account` is byte-identical to before, and configs
+  with zero or one account keep the unchanged `~/.cache/ai-usagebar/anthropic/`
+  cache path (no migration). Per-account TUI tabs remain a follow-up.
 
 ## [0.9.0] — 2026-07-04
 
--- a/README.md
+++ b/README.md
@@ -281,11 +281,50 @@ credentials file and its own cache directory:
   API-key vendors (Z.AI, OpenRouter, DeepSeek) point each module at a
   different key via a wrapper script that sets the env var, plus its own
   `--cache-dir`.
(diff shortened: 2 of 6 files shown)


akitaonrails:
Thanks @zanlucathiago — model contribution. The four back-compat rules held in code, not just prose (the `anthropic_default_creds` split keeping the pre-#14 resolution intact was exactly right), the tests were hermetic by construction, and the scope discipline made this easy to land. Merged as 29f53ed.

Two maintainer adjustments at merge time, both downstream of #15 (which was reported while your PR was open and touches the same file):

1. Your branch landed on top of a new `CredsTarget::{Default, Explicit}` split (2cb7f4b): named-account files now resolve as `Explicit`, which reads **strictly** — on macOS a missing/typo'd account path fails loudly instead of silently reading the *default* account's Keychain item. Your `anthropic_target` tests were kept and now also pin the target kind.
2. Added label validation (the label doubles as a cache subdir nested inside the default cache, so path separators / `..` / `usage.json` are rejected).

Your macOS `collapsible_if` heads-up resolved itself — the #15 rewrite of `write_back` removed that branch entirely. TUI per-account tabs: open a fresh issue when you want to take it, the widget layer you built is the foundation it needs.

zanlucathiago:
Thanks for the thorough review and the merge — glad it landed cleanly.

Both adjustments make sense. The `Explicit` strict-read is clearly the right call (a typo'd account path should fail loud, not silently fall back to the default Keychain item), and label validation is an obvious must once the label becomes a path segment — I should've caught that one myself.

I'll open a separate issue for the per-account TUI tabs and take a crack at it on top of the merged state. Cheers.

Issue #17, opened 2026-07-05
TUI: per-account tabs for Anthropic (follow-up to #14)


zanlucathiago:
Follow-up to #14 — you flagged per-account TUI tabs as the remaining piece, so here's the plan. I'd like to take it.

Right now the TUI is one tab per enabled vendor. The widget/config layer from #14 already knows about `[[anthropic.accounts]]`, but `ai-usagebar-tui` doesn't surface them — you still see a single "Claude" tab for whatever the default account is.

**Proposal:** expand the Anthropic vendor into multiple tabs — the default account (unchanged) plus one tab per `[[anthropic.accounts]]` entry, in config order. Each account tab fetches with its own credentials file (`CredsTarget::Explicit`) and its own `anthropic/<label>` cache — the exact resolution the widget already does, so I'd extract that into one shared helper both call.

Scope I'm proposing:

- Anthropic only. Other vendors are untouched — still one tab each.
- No config changes. It reads the same `[[anthropic.accounts]]` you merged.
- No new keybindings. Tab/h/l already cycle every tab; account tabs just join the list.
- Tab label shows the account, e.g. `Claude` (default) and `Claude · work`. Open to whatever labeling you prefer.
- Settings overlay stays as-is for now (it's per-vendor; editable account arrays are a bigger UI and can be their own follow-up).

Zero behavior change when no extra accounts are configured: same single Claude tab, same order.

If the shape sounds right I'll open a draft PR on top of the merged state. Two things I'd like your read on before I finalize: the tab label format, and whether account tabs should sit right after the default Claude tab (my plan) or grouped some other way.

PR #18, opened 2026-07-06
feat(tui): per-account Anthropic tabs (#17)

zanlucathiago:
Draft for #17 — per-account tabs in the TUI, the follow-up to #14. Rebased on v0.10.0.

### What it does

`ai-usagebar-tui` expands the Anthropic vendor into the default Claude tab plus one tab per `[[anthropic.accounts]]` entry, in config order. Each account tab fetches through its own credentials file (`CredsTarget::Explicit`) and its own `anthropic/<label>` cache. Everything else is one tab per vendor as before.

Here's the two-account layout (sidebar on the wide view):

```
 vendors
 › Claude            ready
   Claude · work     ready
   OpenAI            ready
   Z.AI              error
```

### How

- `TabId { vendor, account }` replaces the bare `VendorId` as a tab's identity; `tabs_from_config` builds the list.
- The account → `(CredsTarget::Explicit, anthropic/<label>` cache`)` mapping is extracted into `AnthropicConfig::account_target`, and the widget's `--account` path (`named_account_target`) now calls it too — so the widget and the TUI resolve accounts through the exact same code, strict-read and all (#15).
- `view.rs` labels an account tab `Claude · work`.

### Scope (matches what we discussed on #17)

- Anthropic only. Other vendors untouched.
- No config changes — reads the same `[[anthropic.accounts]]` from #14.
- No new keybindings — account tabs just join the Tab/`h`/`l` cycle.
- Settings overlay stays per-vendor; editable account arrays would be a separate, bigger UI.
- **Zero change with no extra accounts**: `tabs_from_config` returns exactly `enabled_vendors()`, same order — pinned by a test.

### Two open questions from the issue

1. Tab label format — I went with `Claude · work`. Easy to change.
2. Account tabs sit right after the default Claude tab. Happy to group differently if you'd rather.

### Tests / gates

New tests cover tab expansion + order, the no-accounts == `enabled_vendors()` invariant, `select_primary` landing on the default account tab, and `account_target` resolution. `cargo test` green (261 lib), `clippy -D warnings` clean, `fmt` clean. Also smoke-tested `--account` end to end against a real captured credential (fetch + token refresh + isolated cache). README + CHANGELOG (`[Unreleased]`) updated.

Left as **draft** for your read on the two questions above before I call it final.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,15 @@ Each release is also published at
 
 ## [Unreleased]
 
-Nothing yet.
+### Added
+
+- **Per-account tabs in the TUI** (#17, follow-up to #14). `ai-usagebar-tui`
+  now shows the default Claude tab plus one tab per `[[anthropic.accounts]]`
+  entry, each fetching with its own credentials file and `anthropic/<label>`
+  cache (the same resolution the widget's `--account` uses, extracted into a
+  shared `AnthropicConfig::account_target`). Anthropic-only; other vendors are
+  still one tab each. With no extra accounts configured the tab set and order
+  are unchanged.
 
 ## [0.10.0] — 2026-07-05
 
--- a/README.md
+++ b/README.md
@@ -281,9 +281,9 @@ credentials file and its own cache directory:
   API-key vendors (Z.AI, OpenRouter, DeepSeek) point each module at a
   different key via a wrapper script that sets the env var, plus its own
   `--cache-dir`.
-- The TUI still shows one tab per vendor (not per account); per-account TUI
-  tabs are tracked in
-  [#14](https://github.com/akitaonrails/ai-usagebar/issues/14).
+- The TUI shows the default Claude tab plus one tab per configured
+  `[[anthropic.accounts]]` entry (see the config example below); Tab / `h` / `l`
+  cycle through them like any other tab.
 - On macOS, the login Keychain can hold only one Claude credential per OS
   user, so additional accounts must be file-based as shown above.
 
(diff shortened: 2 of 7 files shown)


akitaonrails:
Merged as 00ecf6e — thanks @zanlucathiago, three for three. The `account_target` extraction was the best change in the PR: the widget's `--account` path now consumes the same resolution the TUI does, so the two front doors can't drift, and named accounts inherit the strict-read semantics from #15 for free. The back-compat pin (`tabs_without_accounts_are_just_enabled_vendors`) and hermetic tests made this a fast merge.

Your two open questions, settled as proposed:
1. **`Claude · work`** — kept. Matches the `·` separator idiom already used in the header and widget formats.
2. **Account tabs directly after the default Claude tab, config order** — kept. Least surprising, and it's what `select_primary` needs anyway.

One maintainer adjustment folded in at merge time: the Settings `SavedAndClose` path now rebuilds the tab set from the reloaded config (new `App::set_tabs` — resets tab states, clamps the selection if the list shrank), so an account added to `config.toml` while the TUI is open shows up without a restart. That gap predated your PR — the old vendor list wasn't rebuilt either — but your `tabs_from_config` made the fix a one-liner, so it rode along with a regression test.

That completes the #14 arc: supported flags (v0.9.0) → config-driven accounts + `--account` (v0.10.0) → per-account TUI tabs (next release). Genuinely great contributions throughout.

zanlucathiago:
Thanks — glad the extraction paid off, that was the part I most wanted to get right so the two paths couldn't drift.

Nice catch on the `set_tabs` rebuild too; adding an account without restarting the TUI is exactly the kind of thing you only notice once it's in front of you. Good to see it ride along with a test.

Fun little arc to work through with you — thanks for the reviews and the quick releases.

PR #19, opened 2026-07-08
feat(anthropic): render model-scoped weekly limits (Fable) from limits[]

EaeDave:
## Problem

Anthropic now enforces per-model weekly caps (currently the **Fable** weekly limit). The usage endpoint reports them **only inside the newer `limits[]` array** — there is no dedicated `seven_day_<model>` field:

```json
{
  "kind": "weekly_scoped", "group": "weekly",
  "percent": 84, "severity": "warning",
  "resets_at": "2026-07-10T10:59:59.628649+00:00",
  "scope": { "model": { "id": null, "display_name": "Fable" }, "surface": null },
  "is_active": true
}
```

(real response captured 2026-07-08 on a Claude Pro account)

Since `UsageResponse` only parses `five_hour` / `seven_day` / `seven_day_sonnet` / `extra_usage`, the widget silently drops this: my Fable week sat at **84% (severity `warning`)** while the tooltip showed only the 55% overall weekly and the bar class stayed `Mid`.

## Change

- `anthropic/types.rs`: parse `limits[]` (lossy, `serde(default)` like the rest of the wire types); lift `kind == "weekly_scoped"` entries with a `scope.model.display_name` into the snapshot.
- `usage.rs`: new `AnthropicSnapshot::scoped: Vec<ScopedWindow>` (API-provided label + weekly window); scoped windows participate in `anthropic_severity` (max + at-cap promotion), so the bar class reflects a scoped cap.
- `widget/render.rs`: tooltip renders one block per scoped window ("Fable weekly", bar, %, reset countdown), between Weekly and Sonnet.
- `tui/panels.rs`: one row per scoped window ("Fable (7d)") after the Sonnet row.
- Labels come from the API, so future scoped models show up without a code change. Accounts without scoped limits render byte-identical to before (covered by existing tests).

`--format` placeholders are deliberately left out: scoped labels are dynamic per account, which doesn't map cleanly onto the static `{sonnet_*}` placeholder pattern. Happy to add a `{scoped_*}` family for the first/worst scoped window as a follow-up if you want it.

## Before / after (`--pretty`, same cached response)

```
│   󰃰  Weekly                  │        │   󰃰  Weekly                  │
│   ███████████░░░░░░░░  55% ↓ │        │   ███████████░░░░░░░░  55% ↓ │
│   ⏱  Resets in 1d 14h        │        │   ⏱  Resets in 1d 14h        │
                                        │                              │
         (nothing)                      │   󰆧  Fable weekly            │
                                        │   ████████████████░░░░  84%  │
                                        │   ⏱  Resets in 1d 14h        │
class: Mid                              class: High
```

## Testing

- `cargo test --release`: 267 passed, 0 failed (adds `parses_weekly_scoped_limits` with the real wire shape above, and `missing_limits_array_yields_empty_scoped`).
- `cargo clippy --release --all-targets`: clean.
- Smoke-tested the built widget against a live Claude Pro account with an active Fable weekly cap (screenshot rendering above).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,16 @@ Each release is also published at
 
 ## [Unreleased]
 
-Nothing yet.
+### Added
+
+- **Model-scoped weekly limits (e.g. the Fable weekly cap)** now render in the
+  widget tooltip, the TUI Claude tab, and the bar's severity class. Anthropic's
+  usage endpoint reports these only inside the newer `limits[]` array
+  (`kind == "weekly_scoped"`, labeled by `scope.model.display_name`) — there is
+  no dedicated `seven_day_<model>` field — so they were previously invisible:
+  a Fable week at 84%/warning showed nothing while the bar stayed green on a
+  55% overall weekly. Labels come from the API, so future scoped models show
+  up without a code change. Accounts without scoped limits are unchanged.
 
 ## [0.11.0] — 2026-07-06
 
--- a/README.md
+++ b/README.md
@@ -352,7 +352,7 @@ Then `hyprctl reload` (no logout needed).
 
 | Vendor | Endpoint | What you see |
 |---|---|---|
-| **Anthropic** | `api.anthropic.com/api/oauth/usage` (undocumented) | Session (5h), Weekly (7d), Sonnet (7d), Extra usage $ |
+| **Anthropic** | `api.anthropic.com/api/oauth/usage` (undocumented) | Session (5h), Weekly (7d), Sonnet (7d), model-scoped weekly (e.g. Fable), Extra usage $ |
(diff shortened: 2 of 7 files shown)


akitaonrails:
Merged as cf64187, shipping in v0.12.0 today — thanks @EaeDave. The dated real-response capture in the regression test is exactly the smoke-discipline pattern this repo runs on, and labeling from `scope.model.display_name` means the next scoped model costs zero code. Verified live before merging: my own Max account has an active Fable weekly cap, and your branch renders the block that v0.11.0 silently drops.

Three small maintainer adjustments folded in: rustfmt on the TUI panels row, the tooltip's scoped block moved after Sonnet so both surfaces render Weekly → Sonnet → scoped in the same order, and two severity regression tests (scoped 84% escalating past a 55% weekly, and scoped-at-100% triggering the extra-usage promotion) — that promotion path was your PR's one behavior change without direct coverage.

On the `{scoped_*}` placeholder family: agreed it's a follow-up, and worth an issue if you want to spec it — first/worst scoped window as static placeholders seems right.

PR #23, opened 2026-07-19
feat: add Kilo, Novita, Moonshot, Grok account-balance vendors

samirhvbr:
## Summary

Adds four **account-balance vendors** that read the remaining credit via each provider's API and render it as money, alongside the existing usage/quota vendors. All are opt-in (disabled until a key is configured) and wired into the Waybar widget, the TUI panels, and the settings overlay.

| Vendor | Endpoint | Unit | Notes |
|---|---|---|---|
| **Kilo** | `GET api.kilo.ai/api/profile/balance` | USD | optional `organization_id` for a team balance |
| **Novita** | `GET api.novita.ai/openapi/v1/billing/balance/detail` | 1/10000 USD | string amounts, divided to dollars |
| **Moonshot** | `GET api.moonshot.ai\|.cn/v1/users/me/balance` | USD / CNY | currency by region (`.ai` = USD, `.cn` = CNY) |
| **Grok (xAI)** | `GET management-api.x.ai/v1/billing/teams/{team}/prepaid/balance` | USD (cents) | see below |

Each endpoint was confirmed against the provider's own docs / open-source client before implementing.

### Grok uses a separate Management (billing) API

xAI does not expose the balance through the inference API. It lives behind a dedicated **Management API** (`management-api.x.ai`) and requires a **Management key** — a *different* key from the inference key, created in the xAI console with billing-read access. The team id is auto-resolved from the key via `/auth/management-keys/validation` when not configured, and `prepaid/balance` returns an inverted-ledger `total.val` in USD cents (negative = credit remaining), converted to dollars. 404 (no prepaid team) is handled.

## Config

```toml
[kilo]
enabled = true
api_key_env = "KILO_API_KEY"      # or inline api_key
# organization_id = "org_..."      # optional: a team balance instead of personal

[novita]
enabled = true
api_key_env = "NOVITA_API_KEY"

[moonshot]
enabled = true
api_key_env = "MOONSHOT_API_KEY"
region = "global"                  # "global" = api.moonshot.ai (USD); "cn" = .cn (CNY)

[grok]
enabled = true
api_key_env = "XAI_MANAGEMENT_KEY" # the MANAGEMENT key, not the inference key
# team_id = "..."                  # optional; auto-resolved otherwise
```

## Notes

- The settings overlay (`s`) was refactored to a single `KEY_VENDORS` table so every key vendor (existing + new) has a field, driven by one list instead of hardcoded blocks. Happy to split that into its own PR or drop it if you'd rather keep this change minimal.
- No changes to the macOS menu bar or GNOME desktop surfaces — these vendors are widget + TUI only for now (like Kimi).
- README (vendor matrix + placeholders) and CHANGELOG updated.

## Verification

- `cargo test --all-targets` — **494 unit tests + 3 integration tests passing** (5 live tests intentionally ignored).
- `cargo fmt --check`, `cargo clippy --all-targets -- -D warnings`, GNOME marker tests, `cargo machete`, and `git diff --check` — clean.
- Audit integration: all four fetchers use the non-blocking cache lock, reject fallback payloads older than 7 days, cap response bodies at 2 MiB, and propagate cache serialization errors.


samirhvbr:
Thanks for the review — every point checked out, including the two Settings regressions I had introduced against `main`. Pushed as `9c0b107`.

**Restore the Settings behavior already present on main**
- `primary_choices` is back on `SettingsState`, computed from `cfg.enabled_vendors()`, with main's fallback semantics (a configured-but-disabled primary displays the first enabled vendor; with none enabled, nothing is persisted). `handle_primary` cycles that list instead of `VendorId::all()`, and `save_to_path` only writes `ui.primary` when `primary_choices.contains(&primary)`.
- The dirty-empty removal is restored via an `update_key` helper matching main's, extended for the key table: dirty + empty removes `api_key`; dirty + non-empty writes it *and* sets `enabled = true`.

**Make the new balance parsers strict**
- Dropped `#[serde(default)]` from the required monetary fields on Kilo, Novita, Moonshot and Grok, so `{}` and 200 error envelopes are schema errors rather than zero-balance snapshots.
- String-encoded amounts (Novita's 1/10000-USD fields, Grok's `total.val`) now fail on empty/non-numeric input instead of `unwrap_or(0.0)`; all amounts are checked finite. Shared helpers `usage::{parse_amount, finite_amount}`.
- Moonshot honors its documented in-band indicators via `BalanceEnvelope::check_ok()` — `status: false` or a non-zero `code` is rejected before any field is read as money.
- Cached monetary fields are required on read too; `reuse_cache` now returns `Result` rather than falling back to a zeroed snapshot.

**Handle xAI management-key scope correctly**
- `Validation` models `scope`. `scopeId` is only used as a team id for `SCOPE_TEAM` (or a legacy response with no `scope`, where `teamId` is the fallback). For `SCOPE_ORGANIZATION` it is **not** adopted as a team — the vendor returns an actionable error asking for `[grok] team_id`, unless an explicit `teamId` is present. Unknown scopes are rejected rather than guessed.

**Include target identity in balance caches**
- Each payload records the target it was fetched for and is rejected on mismatch: Kilo `org:<id>` vs `personal`, Moonshot `<endpoint>|<currency>`, Grok the team.
- One design note worth flagging: for Grok I keyed the cache on the *inputs* (configured `team_id`, else a `DefaultHasher` fingerprint of the management key) rather than the resolved team. Keying on the resolved team would have forced a `/validation` round-trip on every poll, including cache hits. There's a `fresh_cache_makes_no_network_call` test pinning that. The fingerprint is only a change detector — if its value ever shifts, the cost is one extra refetch.

**Finish the setup documentation**
- README: all four vendors in the authentication table, a new "Enabling a vendor" section, the Grok team-scope note, the full env-var list in the resolution order, and config examples.
- `config.example.toml` gained the four sections; it had none before.
- Fixed the contradiction you spotted — the file said "the Settings overlay ... does not enable this vendor" while `save_to_path` sets `enabled = true`. Both README and the example now describe what the code does.
- Added `config_example_toml_parses_and_documents_every_vendor`, which parses the shipped example with the real `Config` struct and asserts every `VendorId` has a section, so this class of drift fails the build rather than the review.

**Two things beyond what you listed**, both because the fixes above were unreachable otherwise:
- When a fetch fails and no *usable* cache exists, the original error is now returned instead of `"<vendor>: no usable cache"`. Without it the new organization-scoped-key guidance never reached the user. This is the same defect you flagged as item 4 on #24, so I fixed it consistently across all four vendors here.
- `cargo fmt` reformatted parts of `kilo/vendor.rs`, `moonshot/vendor.rs` and `novita/vendor.rs` — the branch wasn't fmt-clean. No semantic change; `git diff -w` on those files is empty apart from line breaks.

**Tests** — new coverage for malformed 200 bodies, corrupt/unattributable caches, target changes (organization, region, team), organization-scoped keys, the no-network-call cache hit, disabled-primary selection and save, clearing an inline key, and leaving an untouched key alone. Gate: 411 tests green, `cargo clippy --all-targets -- -D warnings` clean, `cargo fmt --check` clean.

#24 is rebased on top of this and updated separately.

akitaonrails:
Re-audit after rebasing onto current `main` is complete. The author fixes in `9c0b107` are present and correct: strict monetary parsing, target-aware caches, xAI scope handling, enabled-only primary selection, clearing inline keys, and setup documentation all remain covered.\n\nI added `9825511` for compatibility with the safeguards merged in #27: all four new fetchers now acquire cache locks asynchronously, refuse fallback data older than 7 days, cap HTTP bodies at 2 MiB, and propagate cache serialization failures. I also preserved current-main Settings behavior (global Ctrl-C quit, full modifier-chord filtering, resolved config path, and refusal to overwrite unreadable config files) while resolving the rebase.\n\nLocal gate: 494 unit tests + 3 integration tests pass; 5 credentialed live tests are intentionally ignored. Formatting, denied-warning clippy, GNOME marker tests, cargo-machete, and diff checks are clean. Waiting for the rebased three-OS CI before merge.

akitaonrails (approved):
Re-audited on current main. The author addressed every requested change; the rebase preserves current Settings behavior and adds the shared async-lock, bounded-staleness, response-size, and serialization safeguards. Local full gate and Ubuntu/macOS/Windows CI are green.

PR #24, opened 2026-07-19
feat: Anthropic (API) vendor — month-to-date spend + spend-vs-limit

samirhvbr:
## What

Adds an **Anthropic (API)** vendor for Anthropic API/Console organizations, distinct from the existing Claude Code OAuth vendor.

| Vendor | Account and metric |
|---|---|
| **Anthropic** | Claude Code plan usage via OAuth |
| **Anthropic (API)** | API/Console month-to-date spend via an Admin key |

## Authentication

This vendor requires a Console **Admin key** (`sk-ant-admin01-…`) sent in the `x-api-key` header. It is not an inference API key or an OAuth credential. Admin keys are available for organization accounts through Console → Settings → Organization.

## Display

- With a monthly limit: `$1.34 / $1000 · 0%`
- Without a monthly limit: `$1.34/mo`

The prepaid credit balance is not exposed by the API, so the vendor reports spend. `monthly_limit` is a local positive, finite configuration value used only for percentage/severity display.

## How it works

- Fetches the current UTC calendar month from `GET /v1/organizations/cost_report`.
- Follows `has_more` / `next_page` pagination and sums USD amounts reported in the lowest currency unit.
- Validates every result as USD and rejects malformed, non-finite, or non-positive configuration values.
- Uses the shared bounded-response, async cache-lock, and maximum-stale fallback behavior.
- Keys cached results to the Admin credential billing context and UTC month, preventing reuse across organizations or keys.

## Config

Opt-in: the vendor is disabled until enabled and supplied a key through the Settings overlay, an environment variable, or inline configuration.

```toml
[anthropic_api]
enabled = true
api_key_env = "ANTHROPIC_ADMIN_KEY"
# api_key = "sk-ant-admin01-..."       # inline fallback; chmod 600
# monthly_limit = 1000                 # must be positive and finite
```

## Scope

Includes the provider module and registration across configuration, vendor snapshots, CLI parsing, widget/TUI rendering, settings, documentation, and changelog. Desktop-specific GNOME/macOS/Windows surfaces are outside this PR, matching the other opt-in balance vendors.

## Verification

- `cargo test --all-targets`: 517 unit tests and 3 integration tests passed; 5 credentialed live tests ignored.
- `cargo clippy --all-targets -- -D warnings`: clean.
- `cargo fmt --check`: clean.
- GNOME marker check, `cargo machete`, and `git diff --check`: clean.

The implementation was rebased onto the merged balance-vendor foundation before final audit.


akitaonrails (changes requested):
Thanks for the contribution. The Anthropic Admin API endpoint, Admin-key authentication, cents-to-USD conversion, and the general spend-vs-limit approach all check out. I am requesting changes because the current implementation can silently report incorrect billing data, cache it as authoritative, and includes regressions from the stacked #23 branch.

## Please fix in #24

1. **Reject malformed cost responses instead of treating them as $0.** `CostReport`, `Bucket`, and `CostResult` currently use `#[serde(default)]`, and the test explicitly accepts `{}` as a valid zero-cost report. A 200 error envelope or schema change would therefore be cached as genuine zero spend. Require the documented top-level and monetary fields, validate finite numeric amounts, and keep a legitimate `data: []` response as the zero-cost case.

2. **Key the spend cache to the UTC calendar month.** The live query starts at the current month, but the cache stores only `spent` and `limit`. At month rollover, the previous month can be displayed as the current month, including for up to the stale-cache window during an outage. Store the month start in the payload and reject fresh/stale reuse when it differs.

3. **Fail incomplete pagination.** `has_more: true` without `next_page`, a repeated cursor, or reaching `MAX_PAGES` currently returns and caches the partial total. These cases should produce a schema/pagination error.

4. **Preserve the original error when there is no cache.** A first-run 401/403 or schema error is replaced with `anthropic-api: no usable cache`, so the user never sees the actionable Admin-key guidance. Return the original error if fallback data does not exist.

5. **Document the Cost API limitation.** Anthropic states that Priority Tier costs are omitted from this endpoint, so the UI/README should not imply the displayed number is complete spend for affected organizations: https://platform.claude.com/docs/en/manage-claude/usage-cost-api

## Please fix in the stacked #23 dependency before rebasing #24

1. **Restore the Settings behavior already present on `main`.** The refactor removed enabled-only primary choices and always writes any selected vendor as primary. It also ignores a dirty empty API-key field, so clearing an existing inline secret does not remove it. Preserve the existing `primary_choices` semantics and the dirty-empty removal behavior while extending the key table.

2. **Make the new balance parsers strict.** Novita, Moonshot, and Grok default required monetary fields to zero; Moonshot also ignores documented `code` and `status` failure indicators. Malformed 200 responses must be schema errors, not fresh zero-balance snapshots. Cached monetary fields should likewise be required rather than `unwrap_or(0.0)`.

3. **Handle xAI management-key scope correctly.** Validation can return `SCOPE_TEAM` or `SCOPE_ORGANIZATION`; `scopeId` is currently always used as a team ID. Parse `scope` and only auto-use `scopeId` for team-scoped keys. For organization-scoped keys, require or correctly resolve a team. Official schema: https://docs.x.ai/developers/rest-api-reference/management/auth

4. **Include target identity in balance caches.** Kilo organization, Moonshot region/currency, and Grok team changes can currently reuse data cached for the previous target. Record and validate the relevant identity before cache reuse.

5. **Finish the setup documentation.** The authentication/config examples omit most new vendors and currently contradict the code about whether saving a key enables a vendor.

Please add focused tests for malformed 200 bodies, month rollover, incomplete/repeated pagination, target changes, organization-scoped xAI keys, disabled-primary selection, and clearing an inline key. Once #23 is fixed and merged, rebase #24 and I will review the smaller Anthropic-specific delta again.

samirhvbr:
All five points addressed; pushed as `faeccf9`, rebased on the fixed #23. The Anthropic-specific delta is the single commit on top.

**Reject malformed cost responses instead of treating them as $0.00**
- `CostReport.data` / `has_more`, `Bucket.results` and `CostResult.amount` are now required. `{}`, a 200 error envelope, and drifted shapes all fail to deserialize rather than producing an empty report that displays and caches a fabricated zero month. Only `next_page` stays optional, since the docs return it as `null` on the last page.
- Amounts must be numeric and finite; the running total is checked too (`"inf"` parses as `f64::INFINITY` otherwise).
- The old `empty_report_is_zero` test that asserted `{}` was a valid zero-cost report is gone. `a_month_with_no_usage_is_genuinely_zero` replaces it and pins the legitimate case: `data: []`, and a bucket with `results: []`, are still real zeros.

**Key the spend cache to the UTC calendar month**
- The payload records the month it covers and is rejected on mismatch. This also applies on the stale-fallback path — during an outage with only last month cached, the error is surfaced rather than last month's total being displayed as the current one, which was the more dangerous half of what you described.
- Added `fetch_snapshot_at`, a clock seam, so rollover is tested at fixed instants instead of against the wall clock (matching the hermetic-test rule in CLAUDE.md). `fetch_snapshot` is unchanged for callers.

**Fail incomplete pagination**
- `has_more: true` with no `next_page`, a repeated cursor, and exhausting `MAX_PAGES` are each errors now. Previously all three fell into `_ => break` and returned the partial total, which is indistinguishable from a genuinely smaller spend once cached. The repeated-cursor case was the worst of the three: a server echoing one cursor would have summed the same page 12 times into an inflated total.

**Preserve the original error when there is no cache**
- `fallback_with_error` and `fallback_silent` return the original error when nothing usable is cached, so a first-run 401/403 or schema error reaches the user with its Admin-key guidance intact instead of `"anthropic-api: no usable cache"`. Same fix applied to the four #23 vendors.

**Document the Cost API limitation**
- The Priority Tier omission is now stated in the widget tooltip, the TUI panel, the README (endpoint table and placeholder section), `config.example.toml`, and the module docs — with your link. The README wording is explicit that an organization on Priority Tier is seeing **less** than its true total spend, so nothing implies the figure is complete.

**Tests** — malformed 200 bodies and each drifted-envelope shape, the legitimate empty month end-to-end, month rollover both with a fresh cache and during an outage, first-run auth failure preserving the 401, and each pagination failure mode. 18 tests in the module; 426 in the suite. `cargo clippy --all-targets -- -D warnings` and `cargo fmt --check` clean.

The stacking note from the original description still holds: this targets `main`, so #23's commits appear here until it merges.

akitaonrails:
Re-audit completed on the rebased branch.

The author fixes requested in the earlier review are present: strict cost parsing, UTC-month cache handling, pagination, preserved provider errors, and Priority Tier scope documentation.

The final audit commit adds the remaining integration and safety work:
- binds cache entries to the Admin-key billing context so switching organizations cannot reuse another organization’s spend;
- adopts the shared async cache lock, capped response body, and maximum-stale fallback contracts;
- validates USD currency, pagination tokens, accumulated totals, and positive finite monthly limits;
- propagates cache serialization failures;
- updates current-main settings/CLI/config coverage and documentation.

Local verification at `66c1d77446457a65ab4b8dbe72d55a1516f292c0`: 517 unit tests + 3 integration tests passed, 5 credentialed live tests ignored; fmt, clippy with warnings denied, GNOME marker check, cargo-machete, and diff check all passed. Waiting for the fresh three-platform CI run before approval and merge.

akitaonrails (approved):
Approved after re-audit. The earlier requested fixes are present, the remaining current-main integration and cache-isolation issues were corrected in the final audit commit, and Linux/macOS/Windows CI all pass on head 66c1d77446457a65ab4b8dbe72d55a1516f292c0.

PR #26, opened 2026-07-20
fix: data integrity, credential and runtime defects in v0.13.0, plus CI/release gates

samirhvbr:
Independent of #23 and #24 — this branches from `main` and fixes defects in the **released** v0.13.0, not in the new vendors.

The review you left on #23/#24 was about one principle: never display or cache a number the code cannot actually vouch for. Applying that lens to the rest of the codebase turned up the same defect class in shipped vendors, plus several ways the tool can freeze or lose credentials. Every finding below was verified against the `v0.13.0` tag before being fixed, and every fix has a test.

Eight commits, each self-contained and separately reviewable. **Happy to split this into three PRs (data integrity / runtime + credentials / CI) if that reviews better — say the word.**

---

### `7bce7bc` — cache contract

- `MAX_STALE` (7 days) was **declared and never referenced**. Every failure path called `maybe_payload()` with no age check, so after weeks offline the bar kept rendering history as current, distinguished only by `⏸`. Adds `fallback_payload(max_stale)`; all 14 fallback sites go through it.
- A corrupt *fresh* cache became a zeroed snapshot in Anthropic, OpenAI, Z.AI, OpenRouter and DeepSeek — "$0.00", "0%", "Unknown plan", presented as current. They now fall through to a live fetch, which is what Kimi already did.
- **Z.AI accepted an in-band failure as usage.** Errors arrive as HTTP 200 with `success: false` / non-200 `code` / `data: null`. That body parsed cleanly and — worse — was written to the cache *before* anything examined it, destroying the last good payload and clearing the recorded error, then rendered as an unknown plan with empty windows. Now validated before caching.

### `f31b683` — config strictness

All five call sites used `Config::load().unwrap_or_default()`, discarding the distinction `load()` carefully makes between "file absent" and "file broken". A syntax error or permission problem silently produced the *default* vendor set. Each caller now handles it per its contract — the widget reports it through the existing `⚠` fallback and still exits 0. Also denies unknown top-level keys so `[openrouer]` fails by name; deliberately section-level only, so a config carrying a field from a future version still loads.

### `45ca361` — OAuth as a transaction

Both vendors persisted refreshed credentials with `let _ = write_back(...)`, commented as best-effort "because the refresh worked". That holds only while the refresh token is unchanged. On **rotation** the new token exists solely in memory: if the write fails, the token on disk is already spent, this run succeeds, and the next one cannot refresh — the user is logged out with no explanation. Now reported and handled as an auth failure. A failed write carrying only a new access token is still ignored, since nothing is lost. Also fixes OpenAI expiry: `expires_at_secs()` read only the `id_token` exp claim and ignored `auth.json`'s explicit `expires_at`, so a refresh with no new `id_token` left the old expired claim in place and every later run refreshed again.

### `f1d6073` — TUI runtime

- `acquire_lock` parks the thread in a sleep loop for up to 15–45s, and the TUI runs `flavor = "current_thread"`. One blocked task took the reactor with it: keyboard, timer and every other vendor's request stopped. A single concurrent widget invocation was enough. Adds `acquire_lock_async`.
- The keyboard branch created a fresh `spawn_blocking(event::poll)` **every** `select!` iteration; when another branch won, that task was not cancelled and orphans raced on `event::read()`, swallowing keypresses. One reader thread now owns stdin.
- Terminal restoration was straight-line code after the loop, skipped by any early return or panic — leaving raw mode, alternate screen, no cursor. Now an RAII guard.

Test holds the lock and proves an unrelated interval keeps firing on the same current-thread runtime.

### `958bac7` — macOS Keychain

`read_raw` mapped **every** `security(1)` failure to "no item", so a locked login Keychain or denied ACL produced "run `claude` to authenticate" while the credentials sat there intact. Only `errSecItemNotFound` (44) now means absent. Also fixes item selection: with `$USER` unset the read selected by service alone while the write passed `-a ""`, so a refresh could create a second item the read would never find.

**Not fixed here, deliberately:** the OAuth JSON is still passed to `security` in argv. Your comment says there is no stdin path — I checked, and that's correct: `add-generic-password -w` with no value prompts on the tty and asks for confirmation, so it cannot be piped. Removing that exposure needs the native Keychain API (a new macOS-only `security-framework` dependency), which felt like your call rather than something to slip into a fix PR.

### `1044dac` — config location

The binary resolved `config.toml` via `ProjectDirs` (macOS: `~/Library/Application Support/`), while the README, the example, `--help`, the GNOME prefs and the macOS menu bar all said `~/.config/`. Nothing made those the same file — so following the docs got you defaults, and the desktop surfaces reported "no key configured" for keys the binary was using. Platform path stays canonical; the legacy path is honored when it doesn't exist; both desktop surfaces check the same pair. **Nothing is moved or rewritten** — that file can hold API keys. GNOME also stops hard-coding `~/.config` and honors `$XDG_CONFIG_HOME`. Separately, `~` in `credentials_path` (the form the README shows) was kept literally by `PathBuf` and resolved to a directory named `~`; now expanded.

### `b178c5d` — desktop vendor-switch races

GNOME dropped any refresh requested while busy, so a vendor change mid-fetch never started one for the new vendor and the old result landed and stayed. `_refreshToken` didn't help — it's only bumped *inside* `_refresh()`, which had already returned. Now queued, with the vendor captured per attempt and stale results discarded. macOS had no coordination at all: timer, Preferences hook and vendor change could each spawn a subprocess and the last to finish won. Now one in flight, generation-tagged, plus a 45s watchdog matching GNOME's.

### `600c2b8` — CI and release gates

PRs ran **Windows only**. Linux — the platform the widget ships on — was first exercised *after* a tag was pushed, and tags are immutable by this project's own rule, so a failure costs a new patch release. `fmt`, `clippy` and `machete` were in the release checklist but ran nowhere.

More seriously, nothing checked that a release publishes what its tag claims. **At the `v0.13.0` tag, both `.SRCINFO` files still declare `0.8.0`** while `Cargo.toml` says `0.13.0` — and the release shipped. The new `verify-version` job (everything depends on it) requires an existing `vX.Y.Z` tag and agreement across `Cargo.toml`, both PKGBUILDs, both `.SRCINFO`s and `CHANGELOG.md`. I ran it against the real v0.13.0 tree: it fails on both `.SRCINFO`s, which is exactly the release it should have blocked. `workflow_dispatch` also stops accepting an arbitrary commit, and `contents: write` is scoped to the one publishing job.

---

**Gate:** 359 tests green, `cargo clippy --all-targets --locked -- -D warnings` clean, `cargo fmt --all --check` clean, GNOME marker tests pass, the Swift menu bar app compiles.

**Behavior changes worth flagging:** a misspelled config section is now an error rather than ignored; a failure with nothing usable cached now surfaces the error instead of week-old numbers. Both are intentional, but they are user-visible.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,8 +1,7 @@
 name: CI
 
-# Native Windows test signal for PRs. The Waybar widget is Wayland-only, so the
-# Linux gate lives in the maintainer's existing flow; this job proves the CLI +
-# TUI build and test on a real windows-latest runner (no WSL).
+# Cross-platform test signal for PRs and main. Linux carries the strict quality
+# gate, macOS also compiles the native menu app, and Windows runs natively.
 on:
   push:
     branches: [main]
@@ -45,7 +44,7 @@ jobs:
         run: node gnome-extension/marker-logic.test.mjs
 
       - name: Unused dependencies
-        uses: bnjbvr/cargo-machete@main
+        uses: bnjbvr/cargo-machete@v0.9.2
 
   macos:
     name: test (macos-latest)
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,7 +13,58 @@ concurrency:
   group: ci-${{ github.ref }}
   cancel-in-progress: true
 
+permissions:
+  contents: read
+
 jobs:
+  # The platform the widget actually ships on. Until this existed, Linux was
+  # only exercised *after* a tag was pushed — by which point the tag is
+  # immutable and a failure costs a new patch release.
+  linux:
+    name: test (ubuntu-latest)
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v5
+
+      - uses: dtolnay/rust-toolchain@stable
+        with:
+          components: clippy, rustfmt
+
+      - uses: Swatinem/rust-cache@v2
+
+      - name: Format
+        run: cargo fmt --all -- --check
+
+      - name: Clippy
+        run: cargo clippy --all-targets --locked -- -D warnings
+
+      - name: Test
+        run: cargo test --all-targets --locked
+
+      - name: Test GNOME marker logic
+        run: node gnome-extension/marker-logic.test.mjs
+
+      - name: Unused dependencies
+        uses: bnjbvr/cargo-machete@main
+
+  macos:
+    name: test (macos-latest)
+    runs-on: macos-latest
+    steps:
+      - uses: actions/checkout@v5
+
+      - uses: dtolnay/rust-toolchain@stable
+
+      - uses: Swatinem/rust-cache@v2
+
+      - name: Test
+        run: cargo test --all-targets --locked
+
+      # The menu bar app is 700+ lines of Swift with no test harness; at least
+      # prove it still compiles, which nothing verified before.
+      - name: Build the menu bar app
+        run: swiftc -O -o /tmp/ai-usagebar-menubar macos/ai-usagebar-menubar.swift
+
   windows:
     name: test (windows-latest)
     runs-on: windows-latest
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -6,7 +6,7 @@ on:
   workflow_dispatch:
     inputs:
       ref:
-        description: "Tag or commit to build (e.g. v0.2.0)"
+        description: "Existing release tag to build (e.g. v0.13.1)"
         required: true
         default: ""
 
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -11,10 +11,67 @@ on:
         default: ""
 
 permissions:
-  contents: write
+  contents: read
 
 jobs:
+  # Nothing verified that the artifacts about to be published carry the version
+  # the tag claims. At v0.13.0 the packaging `.SRCINFO` files still said 0.8.0
+  # while Cargo.toml said 0.13.0, and the release went out anyway. Everything
+  # downstream depends on this job.
+  verify-version:
+    runs-on: ubuntu-latest
+    outputs:
+      version: ${{ steps.check.outputs.version }}
+    steps:
+      - uses: actions/checkout@v5
+        with:
+          ref: ${{ inputs.ref || github.ref }}
+          # Tags are needed to prove a workflow_dispatch ref is a real tag.
+          fetch-depth: 0
+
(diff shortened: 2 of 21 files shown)


samirhvbr:
Correction, pushed as `3ad3bbd`.

While preparing the follow-up I found that `7bce7bc` in this PR **overstated what it did**. It claims the corrupt-fresh-cache path was fixed for "Anthropic, OpenAI, Z.AI, OpenRouter and DeepSeek". Only the last three were.

Anthropic's `reuse_cache` still did `parse_payload(...).unwrap_or_else(|_| empty_snapshot("Unknown"))`, and OpenAI's `reuse` did `.unwrap_or_else(|_| empty(plan_hint))` — both reached from the fresh-cache path. So an unparseable payload was still rendered as an all-zero snapshot (labelled "Unknown" on Anthropic) and served for the rest of the TTL: exactly the fabricated placeholder that commit said it had removed.

`3ad3bbd` fixes it for real. Both now return `Result` and the fresh path falls through to a live fetch, matching the shape the other vendors already use. Two tests added: a fresh-but-truncated payload triggers a refetch in each vendor rather than surfacing zeros.

Worth noting as evidence the fix is complete: removing the fabrication left `empty_snapshot` and `empty` with **no callers at all** — they existed solely to paper over a parse failure. Clippy's dead-code lint caught them and they are deleted.

Sorry for the churn. The rest of the PR is unchanged; gate is still green (361 tests, clippy, fmt, GNOME, Swift).

akitaonrails:
Maintainer re-audit completed. I found and pushed `d2ffef7` with four follow-ups before merge:

- OpenAI refresh expiry now prefers the newly persisted `expires_at`, then the access-token JWT used by current Codex, rather than the stale ID-token claim.
- TUI Settings now saves to `Config::resolved_path()`, so a macOS legacy config is not shadowed by a new partial canonical file.
- The CI dependency check is pinned to `cargo-machete@v0.9.2` instead of tracking `main`.
- The release dispatch description now accurately requires an existing tag.

Local gate: fmt, clippy with warnings denied, 366 non-live tests, cargo-machete, GNOME marker logic, YAML parse, and `git diff --check` all pass. Waiting for the refreshed Linux/macOS/Windows CI before merge.

akitaonrails (approved):
Re-audited the latest revision and the audit follow-up. Data-integrity paths, configuration resolution, async locking, credential expiry, CI pinning, and cross-platform behavior are covered by the full local gate and the green GitHub Actions matrix. Approved for merge.

PR #27, opened 2026-07-20
fix: reject lenient wire values, escape API-controlled text, bound response bodies

samirhvbr:
Rebased onto merged PR #26. The final diff is the original follow-up plus maintainer audit corrections.

## Final scope

- Reject malformed Anthropic, OpenAI, Z.AI, OpenRouter, and DeepSeek wire values instead of caching or rendering fabricated values.
- Keep legitimate fractional OpenAI percentages while requiring exact integer duration and reset counters.
- Identify Z.AI windows by unit and reject unitless, duplicate, incomplete, or out-of-range layouts.
- Validate OAuth token text and expiry values; propagate bounded-body read failures.
- Escape API-controlled OpenRouter labels, then decode exactly one entity layer after native surfaces strip Pango tags.
- Serialize concurrent vendor cycling and only signal Waybar after persistence succeeds.
- Keep refresh timestamps per TUI tab; make Ctrl-C quit from Settings; ignore modifier chords; preserve multiline diagnostics.
- Refuse to overwrite unreadable config files and propagate cache serialization errors.
- Add `vendor_short` to the native protocol so GNOME and macOS suppress meaningless DeepSeek quota bars while retaining its balance header.
- Bound every vendor and OAuth response body, including chunked bodies without `Content-Length`.
- Keep the OpenAI admin-key setting for compatibility but document that it is currently reserved and inert.

## Verification

Local gate on the rebased head:

- `cargo fmt --all --check`
- `cargo test --all-targets --locked`: 430 unit and 3 integration tests passed; 5 live tests ignored by design
- GNOME marker and protocol tests passed
- `cargo machete` passed
- `git diff --check` passed

The hosted Ubuntu, macOS including Swift compilation, and Windows jobs are the final merge gate.

### DIFF
--- a/config.example.toml
+++ b/config.example.toml
@@ -22,9 +22,11 @@ enabled = true
 # Path to the Codex CLI OAuth credentials. Leave commented out to use
 # the default ~/.codex/auth.json.
 # codex_auth_path = "/home/you/.codex/auth.json"
-# Reserved for a future API-key-only fallback path (admin key reads
-# `/v1/organization/costs`). Today, OpenAI requires Codex OAuth.
-admin_key_env = "OPENAI_ADMIN_KEY"
+# RESERVED — setting this does nothing today. It names the env var a future
+# API-key-only path would read (admin key → `/v1/organization/costs`). No code
+# consumes it: OpenAI usage comes only from Codex OAuth. Left commented out so
+# it isn't mistaken for a working alternative to `codex login`.
+# admin_key_env = "OPENAI_ADMIN_KEY"
 
 [zai]
 enabled = true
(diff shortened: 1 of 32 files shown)


akitaonrails:
Maintainer re-audit completed and pushed in `f783f99` after rebasing onto merged #26. The PR description now records every correction and the full local gate. I will approve and merge only after the new three-OS Actions matrix is green.

akitaonrails (approved):
Re-audited after rebasing onto merged #26. The remaining parser, OAuth, cache, Settings, native protocol, and response-boundary gaps were corrected in f783f99. Local full gate and the fresh Ubuntu/macOS/Windows matrix are green. Approved for merge.

Issue #25, opened 2026-07-19
TUI: local Claude Code context-window monitor


alexzerabr:
ai-usagebar answers "how much of my plan have I used" — session %, weekly %, resets. I'd like to add the local counterpart for Claude Code users: **how full is the context window of the session I'm working in right now**, i.e. how close it is to auto-compaction. I'd like to take it, if you want it here at all.

That signal isn't in the OAuth usage API. It's derivable locally from the transcripts Claude Code writes to `~/.claude/projects/<encoded-cwd>/<uuid>.jsonl`: for each session, the context size at the last turn is the last `type:"assistant"` line's `message.usage.input_tokens + cache_creation_input_tokens + cache_read_input_tokens`. Subagent transcripts (`*/subagents/*`) are sidechains and get skipped.

**Proposal:** a TUI overlay on `c`, modeled on the existing Settings overlay (`App.context: Option<ContextState>`, `from_config` / `handle_key -> Action` / `render`). It lists every session, most-recently-active first, each with a gauge colored by the existing low/mid/high/critical tiers. `Enter` drills into one session on its own — I'd build that detail view out of the same `panels::Section` values the vendor tabs use, so it inherits their gauge, spacing and colors rather than inventing a second look. `↑`/`↓` moves between sessions without leaving the detail view, `Esc` backs out to the list, `r` re-scans.

**Why I'd keep it out of `VendorId` — and out of `TabId::account` too.** The obvious reading is "it's a seventh vendor", and I don't think that survives contact. But the interesting comparison is `TabId::account`, since #14/#17 already solved "one vendor, many instances that need selecting". Both `VendorId` and the account list are *declared in config.toml* and fixed for the process. Sessions aren't: they're discovered at runtime and churn constantly. A tab-per-session TUI would rebuild the tab set on every scan and fight `tab_generation` invalidation, and `active.rs` cycling the Waybar bar through N ephemeral sessions doesn't mean anything. It's the cardinality lifetime that disqualifies it, not the absence of an HTTP client — `is_enabled`, `enabled_vendors` and `Cache` are generic plumbing that would work fine without a network.

Scope I'm proposing:

[...]

PR #29, opened 2026-07-20
feat(context): dockable layouts for the context overlay

alexzerabr:
## What

A small follow-up to the local context monitor (#25 / #28): let the overlay **dock into the dashboard body** instead of floating, and let `v` cycle where it sits.

- **full** — takes the whole body, its own screen (the default)
- **split** — beside the vendor panel
- **bottom** — below it

`[context] layout` sets the mode it opens with; `v` cycles at runtime and the current mode shows in the list's status line.

## Why

The overlay currently floats as a centered 90%×90% modal, so the vendor dashboard bleeds around its edges and the session list gets only a slice of the vertical space. A docked `full` view reads as its own screen; `split` / `bottom` keep the vendor panel and the sessions visible at once. It's your call on the presentation you just shipped — if you'd rather keep the floating modal, or want only `full` without `split`/`bottom`, I'm glad to adjust or close.

## How

- `context::render` now fills the `Rect` it's given as a titled, focus-bordered panel rather than centering a modal (the `centered_rect` helper is gone).
- `view::draw_body` places it per layout — 50/50 split of the dashboard body for `split` / `bottom`, the whole body for `full`. Settings still floats on top as before.
- `ContextLayout` is a plain config enum with a `next()` cycle; `ContextState` carries the current layout and `handle_key` maps `v` to it.
- Additive and back-compatible: `[context] layout` is a new optional field defaulting to `full`, so existing configs and the `enabled = false` default are unchanged. No new dependencies (`Cargo.toml` / `Cargo.lock` untouched), and no edits to `vendor.rs`, `usage.rs`, `widget/run.rs`, `widget/cli.rs`, or `tui/panels.rs`.

## Verification

- `cargo clippy --all-targets --locked -- -D warnings` — clean
- `cargo test --all-targets --locked` — 546 passed (4 new: the `v` layout cycle, the config parse + `full` default, and — against a `TestBackend` buffer — that `full` takes the body while `split` / `bottom` keep the dashboard visible)

Thanks again for ai-usagebar.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,12 @@ Each release is also published at
 
 ## [Unreleased]
 
+### Added
+
+- The local Claude context monitor docks into the dashboard body instead of
+  floating: `v` cycles `full` (its own screen) → `split` (beside the vendor
+  panel) → `bottom`. `[context] layout` sets the one it opens with.
+
 ## [0.14.0] — 2026-07-20
 
 ### Added
--- a/README.md
+++ b/README.md
@@ -545,7 +545,7 @@ make clippy                                        # cargo clippy -D warnings
 - `r` — refresh active tab
 - `R` — refresh all tabs
 - `s` — open Settings overlay (primary vendor + API keys)
-- `c` — open local Claude context sessions (only when `[context] enabled = true`)
+- `c` — open local Claude context sessions (only when `[context] enabled = true`); `v` cycles its layout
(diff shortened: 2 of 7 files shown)


akitaonrails (approved):
Audited the exact merge result: implementation is scoped and backward-compatible; local strict gate and hosted Linux/macOS/Windows CI are green.

PR #31, opened 2026-07-22
fix(anthropic): keep uncapped-plan credit spend visible, in its own currency

### LINKED ISSUE #30: extra_usage block silently dropped when API returns monthly_limit: null — hides credit spend for Pro plans
# Bug: `extra_usage` silently discarded when `monthly_limit` is `null`

## Title

`extra_usage` block silently dropped when API returns `monthly_limit: null` — hides credit spend for Pro plans

## Labels

`bug`, `anthropic`

### Describe the bug

The Anthropic OAuth usage endpoint now returns `"monthly_limit": null` in the `extra_usage` block for plans without a spending cap (e.g. Claude Pro). When this happens, the widget silently discards the entire `extra_usage` block and shows no credit spend information — even though `used_credits` and `currency` are present and valid.

Before this API change, the endpoint returned a numeric `monthly_limit` (e.g. `5000`), so the widget rendered the extra usage bar correctly.

### Root cause

In [`src/anthropic/types.rs:188–193`](https://github.com/akitaonrails/ai-usagebar/blob/main/src/anthropic/types.rs#L188-L193):

```rust
let extra = self.extra_usage.filter(|e| e.is_enabled).and_then(|e| {
    Some(ExtraUsage {
        limit: Cents(e.monthly_limit?),  // ← ? returns None when monthly_limit is null
        spent: Cents(e.used_credits?),
    })
});
```

The `?` operator on `e.monthly_limit` propagates `None` and causes the entire `and_then` closure to return `None`, discarding the block — including the valid `used_credits`.

The `ExtraUsage` struct in [`src/usage.rs:106–109`](https://github.com/akitaonrails/ai-usagebar/blob/main/src/usage.rs#L106-L109) also requires `limit: Cents` (non-optional):

```rust
pub struct ExtraUsage {
    pub limit: Cents,   // ← should be Option<Cents>
    pub spent: Cents,
}
```

### Actual API response (cached)

```json
"extra_usage": {
    "is_enabled": true,
    "monthly_limit": null,
    "used_credits": 14157.0,
    "currency": "BRL",
    "decimal_places": 2,
    "disabled_reason": null
}
```

This represents **R$ 141.57** in credit spend — valid data that should be displayed.

The API also returns a new `"spend"` top-level block with the same information in a more structured format:

```json
"spend": {
    "used": {
        "amount_minor": 14157,
        "currency": "BRL",
        "exponent": 2
    },
    "limit": null,
    "percent": 0,
    "enabled": true,
    "disclaimer": "Usage credits cover you when you hit your plan limits. Learn more"
}
```

### Expected behavior

When `monthly_limit` is `null` but `used_credits` is present:

1. The widget should still display the credit spend (e.g. `R$ 141.57`).
2. The progress bar should either be hidden or show spend without a cap (e.g. just the amount, no percentage).
3. The `{extra_spent}` placeholder should expand to the formatted amount.
4. The `{extra_limit}` placeholder can expand to `—` or empty.
5. The `{extra_pct}` placeholder can expand to `0` or empty.

### Suggested fix

1. Make `limit` optional in `ExtraUsage`:

```rust
pub struct ExtraUsage {
    pub limit: Option<Cents>,
    pub spent: Cents,
}
```

2. Update the conversion in `types.rs`:

```rust
let extra = self.extra_usage.filter(|e| e.is_enabled).and_then(|e| {
    Some(ExtraUsage {
        limit: e.monthly_limit.map(Cents),
        spent: Cents(e.used_credits?),
    })
});
```

3. Update `ExtraUsage::percent()` to return `0` (or `None`) when `limit` is `None`.

4. Update the widget renderer and tooltip to handle the no-limit case — show `$X.XX` without a progress bar or with an unbounded display.

5. Consider also consuming the new `"spend"` top-level block as a fallback or primary source, since it includes `currency` and `exponent` for proper formatting.

### Environment

- **ai-usagebar version**: 0.14.0 (compiled from `main` at `167ceb2`)
- **Platform**: macOS 15 (aarch64)
- **Anthropic plan**: Claude Pro
- **Currency**: BRL

samirhvbr:
Fixes #30.

## The bug

The usage endpoint now sends `extra_usage.monthly_limit: null` for plans with no spending cap (e.g. Claude Pro). The conversion did `Cents(e.monthly_limit?)` inside the `and_then`, so the `?` discarded the **entire block** — including a perfectly valid `used_credits`. The reporter's R$ 141.57 of real credit spend vanished from every surface.

## The semantic argument

The strictness that discards the block came from the `f783f99` re-audit, and its principle is right: an enabled-but-incomplete block must not manufacture $0.00. But it conflated two different inputs:

| Input | Meaning | Correct behaviour |
|---|---|---|
| `monthly_limit: null` + valid `used_credits` | **no spending cap** — semantic, observed on Pro (#30) | show the spend, no invented percentage |
| `used_credits` missing/null | no spend datum | drop the block (nothing truthful to show) — unchanged |

`ExtraUsage.limit` is now `Option<Cents>`. The block is only dropped when `used_credits` itself is unusable. Nothing is fabricated in either direction: the spend shown is exactly what the API sent, still gated by the strict `de_opt_cents`, and with no denominator there is no percentage — the bar and severity stay calm at 0 instead of inventing one.

One deliberate call-out: with `#[serde(default)]` an *absent* `monthly_limit` is indistinguishable from an explicit null at the struct level, so both render spend-only. Distinguishing them would take a double-option deserializer whose only effect is hiding real spend when a secondary field drifts — the exact failure mode of #30. The rendered wording (`Limit: none reported`) is precisely true of both cases and deliberately does **not** say "uncapped plan", which would infer a plan tier the payload doesn't state.

## Per surface

- **widget**: `{extra_spent}` unchanged; `{extra_pct}` → `0`; `{extra_limit}` → `—`, deliberately **non-empty** — GNOME (`extension.js` checks `d.extra.limit` truthiness) and the macOS menu bar (`limit.isEmpty` in `parse`) hide the whole extra row on an empty limit, which would re-hide the spend this fix recovers. No JS/Swift changes needed.
- **tooltip**: spend in bold as before; `Limit: none reported`.
- **TUI panel**: the amount alone (no "of $X.XX"), footnote "no monthly limit reported".

## Currency (the second half of the reporter's payload)

His block says `currency: "BRL"`, and both `currency` and `decimal_places` were ignored — every amount went through `fmt_dollars`, hard-coded `$` and cents. R$ 141.57 would have rendered as **"$141.57"**, a claim about the wrong currency — the same defect class as a fabricated number. Now:

- known codes get their symbol (`$`, `R$`, `€`, `£`, `¥`), mirroring the per-currency match `deepseek::format_money` already established; unknown codes render as `AMOUNT CODE`, still truthful;
- zero-exponent currencies (JPY) format without an invented decimal point — including when `decimal_places` is absent, where a flat default of 2 would show ¥500 as ¥5.00 (the fallback is per-currency);
- payloads without the fields keep the exact historical `$`/cents output, so nothing changes for accounts the old code rendered correctly.

Both new wire fields are gated at the parse boundary like everything else in this file:

- `decimal_places`: integer in 0..=6 (covers all of ISO 4217). **Integral floats are tolerated** for the same reason `de_opt_cents` tolerates them — this endpoint floats its numbers (`used_credits: 14157.0` in the reporter's own payload), and rejecting `2.0` would fail the whole response *after* the fetch has already cached it, evicting the last good payload and leaving a persistent ⚠.
- `currency`: three ASCII uppercase letters. The value is embedded verbatim in Pango bar markup and in the `;;`-delimited desktop FORMAT protocol, so an arbitrary string is an injection vector besides being drift.

## Tests

- The reporter's actual cached fixture (integral-float BRL cents) exercised at all three layers: wire→snapshot, placeholder/tooltip render, TUI panel.
- The BRL assertions live at the **surface** layer on purpose: reverting the currency wiring in `render.rs`/`panels.rs` fails the suite, not just the unit layer.
- `enabled_incomplete_extra_usage_does_not_invent_zero_money`'s third case pinned the null-limit discard — the behaviour #30 demonstrates is wrong — and was replaced; the no-spend cases keep their pin in `enabled_extra_usage_without_spend_is_dropped`.
- Drift gates covered: `decimal_places` 7/-1/100/2.5 rejected, 2.0 accepted; currency `"brl"`/`""`/`"R$"`/`"USD;;0"`/`"<b>"` rejected.

Not consumed here: the new top-level `spend` block from the issue (`amount_minor`/`currency`/`exponent`). It's the cleaner source long-term, but switching the source of truth deserves its own PR — happy to follow up.

**Gate:** 551 tests, `cargo clippy --all-targets --locked -- -D warnings`, `cargo fmt --all --check`, GNOME marker tests, Swift menu bar compiles.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,15 +34,16 @@ Each release is also published at
   `decimal_places` fields were ignored and every amount was formatted as `$`
   with a hard-coded cent scale — the #30 reporter's R$ 141.57 would have shown
   as "$141.57", a claim about the wrong currency. Known codes get their symbol
-  (`R$`, `€`, `£`, `¥`), unknown ones render as `AMOUNT CODE`, zero-exponent
-  currencies (JPY) format without an invented decimal point, and payloads
-  without the fields keep the historical `$`/cents behaviour (JPY defaults to
-  its ISO zero-decimal scale when the field is absent). Both new fields are
-  gated at the parse boundary: `decimal_places` outside 0..=6 is schema drift
-  (integral floats are tolerated, since this endpoint floats its numbers), and
-  `currency` must be a three-letter ISO alpha code — the value is embedded in
-  Pango markup and the desktop `;;` protocol, so an arbitrary string would be
-  an injection vector besides being drift.
+  (`R$`, `€`, `£`, `¥`), unknown ones render as `AMOUNT CODE`, and an explicit
+  exponent is honored exactly, including zero- and three-decimal currencies.
+  If a currency is present but its exponent is absent, the raw value renders as
+  `N minor units CODE` rather than guessing and silently corrupting the amount;
+  payloads with neither field keep the historical `$`/cents behaviour. Both
+  new fields are gated at the parse boundary: `decimal_places` outside 0..=6 is
+  schema drift (integral floats are tolerated, since this endpoint floats its
+  numbers), and `currency` must be a three-letter ISO alpha code — the value is
+  embedded in Pango markup and the desktop `;;` protocol, so an arbitrary
+  string would be an injection vector besides being drift.
 
 ## [0.14.0] — 2026-07-20
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,35 @@ Each release is also published at
   floating: `v` cycles `full` (its own screen) → `split` (beside the vendor
   panel) → `bottom`. `[context] layout` sets the one it opens with.
 
+### Fixed
+
+- **Credit spend is no longer hidden on plans without a spending cap** (#30).
+  The usage endpoint sends `extra_usage.monthly_limit: null` for uncapped
+  plans (e.g. Claude Pro); the whole block was discarded, hiding genuine
+  `used_credits`. A null limit is semantic — "no cap" — not schema drift, so
+  `ExtraUsage.limit` is now optional: the spend renders on every surface, the
+  tooltip says `Limit: none reported` (stating the wire fact rather than
+  inferring a plan tier), the TUI shows the amount without a denominator, and
+  `{extra_limit}` expands to `—` (deliberately non-empty: GNOME and the macOS
+  menu bar hide the whole extra row on an empty limit).
+  The block is still dropped when `used_credits` itself is missing — without
+  the spend there is nothing truthful to show — and no percentage is invented
+  when there is no denominator.
+
+- **Extra usage renders in its own currency.** The block's `currency` and
+  `decimal_places` fields were ignored and every amount was formatted as `$`
+  with a hard-coded cent scale — the #30 reporter's R$ 141.57 would have shown
+  as "$141.57", a claim about the wrong currency. Known codes get their symbol
+  (`R$`, `€`, `£`, `¥`), unknown ones render as `AMOUNT CODE`, zero-exponent
+  currencies (JPY) format without an invented decimal point, and payloads
+  without the fields keep the historical `$`/cents behaviour (JPY defaults to
+  its ISO zero-decimal scale when the field is absent). Both new fields are
+  gated at the parse boundary: `decimal_places` outside 0..=6 is schema drift
+  (integral floats are tolerated, since this endpoint floats its numbers), and
+  `currency` must be a three-letter ISO alpha code — the value is embedded in
+  Pango markup and the desktop `;;` protocol, so an arbitrary string would be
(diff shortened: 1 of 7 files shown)


akitaonrails (approved):
Re-audited exact head 7bf191e. The maintainer amendment preserves a missing decimal_places value instead of guessing a currency exponent, with KRW (zero-decimal) and KWD (three-decimal) regressions. Full local gate and hosted Linux/macOS/Windows CI are green.

PR #32, opened 2026-07-22
fix(tooltip): count XML entities as one glyph; share the window block

4nrry:
Two changes to the shared tooltip primitives, independent of any vendor. First of a three-PR stack; the other two build on it.

## Scope

- `visible_width` measured markup by source characters, so an escaped character inflated the count: `&amp;` scored 5 instead of 1. Every bordered-tooltip row containing one was under-padded and stopped short of the right border, leaving a ragged box. Any vendor whose API-supplied labels can contain `&`, `<` or `>` is affected, since the label is escaped for Pango before the box is measured.
- A named or numeric entity now counts as one glyph. A bare `&` still counts as one character, as before.
- Extract the three-line usage-window block (icon + label, bar + percentage, dim reset countdown) that `zai/vendor.rs` and `openai/vendor.rs` had duplicated byte-for-byte into `tooltip::push_window`, with a new `elapsed` argument feeding the existing `pango::progress_bar` pace marker. Both existing callers pass `None`, so their rendered output is unchanged.

## Verification

Local gate:

- `cargo fmt --check`
- `cargo test`: 550 tests passed (546 on `main`), 5 live tests ignored by design
- `cargo machete` passed
- `git diff --check` passed
- GNOME marker tests passed

New coverage: entity widths (`&amp;`, `&lt;`, `&quot;`, numeric `&#38;`), a bare ampersand and a non-entity `&…;` run, `escape(label)` measuring equal to the raw label for every label shape, and a `render_bordered` regression asserting every line of the box has the same visible width when one row contains an escaped character.

## Limitations

- The extraction is behaviour-preserving for `zai` and `openai` by construction (both pass `None`), so nothing pins their tooltips against a future change to `push_window` beyond their existing tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,13 @@ Each release is also published at
 
 ## [Unreleased]
 
+### Fixed
+
+- Bordered tooltips no longer ragged-edge on rows containing an escaped
+  character: `visible_width` counted `&amp;` as five glyphs instead of one, so
+  every such row stopped short of the right border. Affects any vendor whose
+  API-supplied labels contain `&`, `<` or `>`.
+
 ### Added
 
 - The local Claude context monitor docks into the dashboard body instead of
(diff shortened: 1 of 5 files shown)


akitaonrails (approved):
Re-audited exact head a7657cf. Maintainer amendment resolved the v0.15.0 changelog conflict and restricted entity decoding to valid XML/Pango named and numeric references, with malformed-reference regressions. Full local gate and hosted Linux/macOS/Windows CI are green.

PR #33, opened 2026-07-22
feat(antigravity): Google Antigravity vendor — four quota windows

4nrry:
Adds Google Antigravity as a vendor. Stacked on #32, so its commit appears here until that merges; the Antigravity delta is the single commit on top.

Antigravity reports four independent quota windows: a 5-hour and a weekly limit for each of two model pools — Gemini, and Claude & GPT OSS. The pools do not share a budget, so exhausting one leaves the other untouched.

`ai-usagebar --vendor antigravity --pretty` on a Google AI Pro account:

Antigravity tooltip rendered by the widget

The two 5-hour windows show the same countdown here only because both pools are
idle — an untouched bucket's reset slides with the clock. Gemini's weekly is the
one in use (8%, resetting in 6d 1h) and has anchored, which is why it differs.
That is why the four windows each keep their own countdown rather than being
collapsed into shared rows.

The `⏸` after the bar text is the stale marker: this was captured with no
Antigravity product running, so the figures come from cache. Once a cached
window is past its reset the payload is refused outright rather than shown —
see the cache notes below.

## Scope

- Google ships three Antigravity products (Antigravity 2.0, the `agy` CLI, and the IDE) that all draw on the same account-wide quota, and a machine may run any combination. Rather than preferring one, the fetcher discovers every local server and trusts the first that answers. There are no credentials to configure — the vendor is a plain `enabled` flag.
- Registered across `vendor.rs`, `usage.rs`, `config.rs`, `cli.rs`, `run.rs`, `active.rs` and the TUI, following the existing multi-vendor pattern.

## Endpoint notes, verified against live responses

- Quota comes from `RetrieveUserQuotaSummary`, not `GetUserStatus`. The latter’s per-model `quotaInfo` mirrors whichever bucket is scarcest, so reading it as a window reports the same figures for all four.
- The server binds with `--https_server_port 0`, so its port is drawn from the ephemeral range and is discovered via `/proc` rather than assumed. `ANTIGRAVITY_LS_ADDRESS` overrides discovery.
- Antigravity 2.0 guards the RPC with a CSRF token served at `/`; `agy` serves no such page and answers unauthenticated, so the token is optional rather than required.
- An untouched bucket’s reset slides with the clock and only anchors on first use. Two windows can therefore look synchronised while idle and diverge once either pool is used, so each window keeps its own countdown and they are never merged.
- Percentages are reported as *consumed*, matching every other vendor. Antigravity’s own UI shows the inverse (what remains), so the figures read as complements of each other.

## Data integrity

- A bucket without a finite `remainingFraction` is a schema error, not a reassuring "0% used".
- A cached payload missing a required window is rejected so the caller refetches, rather than rendering zeros for the rest of the TTL.
- The cache fingerprints the signed-in account (hashed, never stored in clear), so switching Google accounts cannot serve the previous account’s figures. Identity is resolved from the same loopback call that supplies the plan name, so verification costs no extra round-trip. With no local server there is nothing to compare against — that cache is served rather than discarded, since nothing is consuming quota while Antigravity is closed.
- A fetch failure with nothing usable cached surfaces the original error, so the "open Antigravity" guidance reaches the user instead of a generic cache miss.
- A cached window whose reset has already passed is refused rather than served as current. `MAX_STALE` is seven days while the session window is five hours, so with nothing running the cache would otherwise report a period that has since rolled over, with its countdown pinned at "now". A window with no reset time is unknown rather than expired and is still served.
- Response bodies are bounded, the async cache lock is used, and fallback data older than `MAX_STALE` is refused.

## Verification

Local gate:

- `cargo fmt --check`
- `cargo test`: 583 tests passed, 5 live tests ignored by design
- `cargo machete` passed
- `git diff --check` passed
- GNOME marker tests passed

36 tests cover the module. The quota fixture is a real `RetrieveUserQuotaSummary` capture (2026-07-22, Antigravity 2.0 build 2.3.1 / `agy` 1.1.5) with percentages edited to distinct non-zero values so a slot mix-up cannot pass. Verified live end to end against both the desktop app and an `agy` session.

## Limitations — please read before merging

- **Server discovery is Linux-only.** It reads `/proc` directly rather than shelling out to `ss`/`lsof`. On macOS and Windows `discover_ls_ports` returns empty and the vendor works only via `ANTIGRAVITY_LS_ADDRESS`. The `/proc` parser test is `#[cfg(target_os = "linux")]`; everything else in the module is portable and runs on all three CI platforms. I did not attempt a macOS implementation because I cannot test one.
- **Quota is only served while an Antigravity product is running.** Unlike Anthropic and OpenAI, which read a persisted OAuth file and call a remote endpoint, Antigravity persists no credential on disk — both binaries link the OS keyring — and the remote CloudCode API exposes no quota RPC that I could find. `RetrieveUserQuotaSummary` belongs to the local `exa.language_server_pb` protocol. With everything closed the widget serves the cache and then reports an actionable error. I could not find a way to match the other vendors’ always-available behaviour without reverse-engineering keyring storage.
- For `agy` this means an **interactive** session; `agy models` opens ports but does not answer the quota RPC.
- The endpoint is undocumented and was derived from the shipped binaries, so it is subject to the same silent-drift risk as the other undocumented vendors. It has no live smoke test in `tests/live.rs` because it needs a running local app rather than a credential.


akitaonrails (approved):
Re-audited after merging current main. The Antigravity parser and cache paths now reject ambiguous, malformed, duplicate, and out-of-range data; response bodies are bounded on every status; unusable fallback data preserves the actionable live failure. Full local gate and hosted Linux/macOS/Windows CI pass at 631b208.

PR #34, opened 2026-07-22
feat(gnome): grouped dropdown and per-pool panel for two-pool vendors

4nrry:
Third of the stack, on top of #32 and #33; their commits appear here until those merge. The GNOME delta is the single commit on top.

Antigravity is the first vendor whose windows come in two independent quota pools, each with its own 5-hour and weekly limit. Rendered as four flat rows they read as two unrelated pairs, so the dropdown groups them.

Grouped dropdown and per-pool panel

*`Session` and `Weekly` headings each holding one bar per pool, and the panel drawing all four segments with the pool initial.*

Panel pools preferences

*The two new preference rows. `Limiar do modo automático` is insensitive unless `Pools no painel` is set to `auto`. Note the bar width in this shot is 4 rather than the default 8 — with `both` selected the panel draws four segments, and that is the trade-off called out under Limitations below.*

*The repo currently bundles no GNOME screenshot — happy to commit these into `gnome-extension/` instead of linking them, if you want them in-tree.*

## Scope

- The dropdown groups the four rows under `Session` and `Weekly` headings, one bar per pool (`Gemini`, `Claude & GPT OSS`). The panel prefixes each segment with the pool’s initial (`G 5h`, `C 7d`).
- New `Panel pools` preference: draw both pools (default), either alone, or `auto`.
- `auto` falls back to the other pool once **either** window of the shown one passes `Auto threshold` (default 95%). Switching on the 5-hour alone would strand the user on a pool whose weekly is what ran out. With both pools spent it stays put rather than flapping.
- Pace markers now render for the extra row too.
- Supported GNOME Shell range raised to 45–50.

## Compatibility

- **The grouped layout is opted into by the data, not by a vendor id.** A vendor that names its primary rows via the new `{session_model}` / `{weekly_model}` placeholders gets it. Single-pool vendors emit neither and are untouched; I verified Anthropic and OpenAI still render `5h` / `7d` with no pool tag against the real binary.
- **New format fields are appended before the sentinel**, so no existing index shifts. A binary predating them echoes the placeholders back, `field()` discards them, and the flat four-row layout is used — pinned by a test asserting `FIELD.vendorShort` is still 16.
- **The slot mapping is unchanged.** The primary pool still fills the generic session/weekly fields, so the panel bar and the `show-session` / `show-weekly` toggles keep working as they do everywhere else. Only the labels and the visual order differ.
- `Panel pools` composes with the existing window toggles rather than duplicating them: turning the weekly bar off leaves one segment per pool, at the width of a single-pool vendor.

## Verification

Local gate:

- `cargo fmt --check`
- `cargo test`: 583 tests passed
- `cargo machete` passed
- `git diff --check` passed
- `node marker-logic.test.mjs` passed; `node --check` on `extension.js`, `prefs.js`, `marker-logic.js`
- `glib-compile-schemas` accepts the schema; both new keys resolve with their documented defaults

New JS coverage: pool tag derivation, tag disambiguation when two pools share an initial, and the `pickPool` boundaries (below threshold, threshold exactly, weekly rather than 5-hour exhausted, both pools spent, missing second pool).

Every `panel-pools` mode was exercised against real binary output outside GNOME Shell, including the non-regression path for Anthropic and OpenAI.

## Limitations — please read before merging

- **The pool tag is a single letter derived from the model-group name** (`G`, `C`). If Google renames a group so both share an initial, `disambiguateTags` widens both until they differ, but the result can get long. Icons would read better; I left them out deliberately, since the panel is a single `St.Label` with Pango markup — Pango cannot embed images, so SVG would mean restructuring it into a `St.BoxLayout` of `St.Icon` children, and shipping vendor logos in an AUR-distributed package raises a trademark question that is yours to decide, not mine.
- **The `both` default widens the panel to four segments.** `bar-width`, `show-bars` / `show-percent`, and turning the weekly bar off all trim it, but the default may be wider than you want on smaller screens — easy to change to `primary` if you prefer.
- Only Antigravity exercises the grouped path today, so the layout is generic by construction but proven by one vendor.
- I did not touch the macOS menu bar app; it keeps the flat layout and is unaffected.


akitaonrails (approved):
Re-audited on current main. Fixed absent/partial secondary pools blanking the panel (including window-toggle combinations), made pool tags Unicode-safe and derive from either available window, and replaced the invalid Antigravity OAuth/npm helper with its real local-product flow. Strict schema/JS checks, full local gate, and hosted Linux/macOS/Windows CI pass at c5dc9cb.

PR #35, opened 2026-07-23
feat(macos): full vendor parity, quick switch submenu, and ring layout

lucasliet:
image

## What changes

Brings the macOS menu bar app to full vendor parity with the Rust binary (all vendors except Antigravity), and adds a compact ring layout, a quick-switch submenu, and the first automated test coverage for the Swift app.

- **Eleven-vendor parity.** The selector previously exposed only five vendors (Anthropic, OpenAI, Z.AI, OpenRouter, DeepSeek); it now covers all binary vendors except Antigravity. Added Kimi, Kilo, Novita, Moonshot, Grok (xAI), and Anthropic (API).
  - **Real balances for every balance-only vendor.** OpenRouter, DeepSeek, Kilo, Novita, Moonshot, Grok, and Anthropic (API) now render their actual balance/credits (`cr <amount>`) via per-vendor format fields, instead of fake 0% session/weekly rows. Anthropic (API) shows a spend-vs-limit bar when a monthly limit is configured. The balance is dispatched by the selected vendor, not by `vendor_short` (which collides between Kimi and Moonshot).
  - **TOML `enabled` handling matches the Rust config.** Bare booleans and inline comments are parsed, and an omitted `[vendor].enabled` reproduces the `src/config.rs` defaults. Both the Preferences picker and the "Trocar vendor" submenu only offer enabled vendors.
  - **Generalized `config.toml` reader.** Reads any key under any `[section]`, so `api_key_env` is resolved per vendor instead of being hardcoded.
- **Quick vendor switch submenu.** A "Trocar vendor" submenu lists only configured/enabled vendors with a checkmark on the active one, so switching no longer requires opening Preferences.
- **Optional ring indicator.** A new "Estilo do indicador" preference selects between the default block bars (░█) and a compact Core Graphics ring. The ring paints usage as a severity-colored arc over a faint track and honors the pace marker the same way block bars do. The track adapts to the effective appearance — faint white on dark menu bars (where the dark `COLOR_EMPTY` would be invisible), `COLOR_EMPTY` on light ones.
- **Appearance-aware status text.** Text colors resolve against the effective status-bar appearance via `menuBarTextColor()`, and the app re-renders immediately when the system appearance changes (KVO on `effectiveAppearance`).
- **Pure-logic test harness.** The single-file app has no Xcode project, so it is compiled with `-D SWIFT_TEST_HARNESS` alongside a test file that calls its helpers directly (`macos/run-tests.sh`). Covers arc geometry, TOML `enabled` parsing, Rust defaults, and per-vendor balance dispatch. The CI macOS job runs it.

## Fixes

- **Ring pace arc.** The overshoot arc restarted at 12 o'clock and overpainted the calm fill; it now spans from the elapsed marker to the current percentage.
- **Preferences crash.** The SwiftUI host is installed through `contentViewController`, avoiding an AppKit exclusivity crash during window measurement.

## Not supported

- **Google Antigravity on macOS.** The binary only discovers its local language server on Linux (via `/proc`); on macOS there is no reachable quota source, so Antigravity is not offered. Safe macOS server discovery (e.g. via `lsof`) is dedicated future work. (Antigravity remains supported in the GNOME extension and the TUI.)

## Why

The macOS surface had drifted from the binary's vendor and placeholder contracts: it assumed hardcoded environment variables, treated balance-only vendors as rate-limit windows (fake 0% rows), and ignored the `enabled` flag for opt-in vendors. This PR closes that gap and adds the compact ring layout (the block bars take too much horizontal space in the menu bar).

## Test plan

- [x] `swiftc -O -parse-as-library -o /tmp/ai-usagebar-menubar macos/ai-usagebar-menubar.swift` (the CI macOS build step)
- [x] `./macos/run-tests.sh` (44 assertions: arc geometry, TOML parsing, Rust defaults, per-vendor balances)
- [x] `cargo fmt --all -- --check`, `cargo clippy --all-targets --locked -- -D warnings`, `cargo test --all-targets --locked` (unchanged; no `src/` or `gnome-extension/` changes)
- [x] `node gnome-extension/marker-logic.test.mjs`
- [x] Manual validation: Preferences opens without the Swift exclusivity crash; the ring preference is reachable and arcs remain colored; disabled vendors stay absent from "Trocar vendor"; appearance changes re-render immediately; balance-only vendors show `cr <amount>` and no 5h/7d rows.

## CHANGELOG

Updated in `CHANGELOG.md` under `[Unreleased]`.

## Notes for reviewers

The macOS app has no XCTest bundle. The compile gate plus the pure-logic harness (`macos/run-tests.sh`, now in CI) cover the changed surface; block bars remain the default layout. The `vendor_short = "kmi"` collision between Kimi and Moonshot is a latent Rust bug (out of scope for this macOS PR) — the macOS parser sidesteps it by dispatching on the selected vendor.


akitaonrails (changes requested):
Thanks for expanding the macOS integration. The direction is useful, and the existing Rust/GNOME gate passes locally, but this needs the following fixes before it is safe to merge:

1. Preserve meaningful output for every added vendor. The Swift FORMAT requests only `{or_balance}`, and the parser only handles `dsk` and `opr` as balance-only vendors. Kilo, Novita, Moonshot, and Grok currently fall through to their generic `session_pct = 0` / `weekly_pct = 0` aliases, so the app renders fake 5h/7d 0% rows and drops their actual balances. Extend the desktop protocol with a generic headline/balance field or explicit vendor fields, render according to `vendor_short`, and cover each supported vendor with parser/render tests. Anthropic API should likewise use labels/data appropriate to monthly spend rather than presenting duplicate session/weekly semantics.

2. Make enabled/disabled handling match the Rust config. `configValueTOML()` only accepts quoted strings, so normal TOML such as `enabled = false` never reaches `configEnabledTOML()`. Missing values are then treated as enabled even though the opt-in vendors default to false in Rust. The Preferences picker is also built from an unconditional list and therefore still offers disabled vendors. Parse bare booleans correctly, reproduce the Rust defaults for omitted sections/keys, and filter both the Preferences picker and quick-switch menu. Please add coverage for explicit false, explicit true, omitted/defaulted vendors, inline comments, and custom `api_key_env`.

3. Resolve the Antigravity macOS reachability gap. The Rust binary discovers local Antigravity ports only on Linux; on non-Linux platforms it returns no discovered candidates unless `ANTIGRAVITY_LS_ADDRESS` is set. Checking for `~/.gemini/*` directories and telling users merely to open the app/IDE/agy therefore marks an unusable setup as configured. Either implement safe macOS server discovery, or require/detect a usable address on macOS and update the UI copy and documentation accordingly.

4. Fix the ring pace arc. `drawArc` always begins at 12 o'clock, so `drawArc(p - eFrac, over)` overlays the start of the calm arc instead of drawing from the elapsed marker to `p`; it also leaves the total visible fill short. Make the helper accept start/end fractions and draw `[0, min(p,e)]` plus `[e,p]`. A testable geometry helper with cases such as p=80/e=50 would prevent regression.

5. Update the user documentation, not only CHANGELOG.md. README.md and macos/README.md still explicitly say the selector supports five vendors and Kimi is unsupported; macos/README.md and macos/INSTALL.md also omit the ring and quick-switch behavior. Please update the scope/version wording, settings table, vendor-specific display behavior, and any Antigravity platform requirement.

6. Run the complete hosted gate after updating the branch. Both current CI attempts are held as `action_required`, so the required macOS Swift compile has not run. A maintainer will need to approve the fork workflow; this should not merge until Linux, macOS, and Windows jobs are green.

Once these are addressed, please leave a note and we can re-audit the updated head.

lucasliet:
Thanks for the thorough review — all six blockers are addressed in [e1ccf7b](https://github.com/akitaonrails/ai-usagebar/pull/35/commits/e1ccf7b0b506a1b075b3ebd923d8dbc69d3c79bf) and https://github.com/akitaonrails/ai-usagebar/pull/35/commits/ab5b1d169178aa434af5b875d8b79fcdce864528 (just pushed). Point by point:

**1. Real balances per vendor (no more fake 0% rows).** The FORMAT now requests each balance-only vendor's own field (`{ds_balance}`, `{kilo_balance}`, `{nv_balance}`, `{km_balance}`, `{grok_balance}`, `{aapi_headline}`), and `parse()` dispatches the balance by the **selected vendor** rather than `vendor_short`. This matters because Kimi and Moonshot both report `vendor_short = "kmi"` in the Rust binary, so keying on `vendor_short` would collide and read the wrong field. Anthropic API renders a spend-vs-limit bar when a monthly limit is configured (`aapi_pct`/`aapi_spent`/`aapi_limit`), otherwise its headline. Balance-only vendors set `hasUsageWindows = false`, so the 5h/7d rows are suppressed. Note: the `kmi` collision is a latent Rust bug, but it's out of scope for this macOS PR — flagging it for a follow-up.

**2. TOML `enabled` matches the Rust config.** `configValueTOML` now parses bare booleans (`enabled = false`), strips inline comments (`enabled = false  # opt-in`), and an omitted `[vendor].enabled` falls back to `defaultEnabled()`, which reproduces the `src/config.rs` defaults (anthropic/openai/zai/openrouter = `true`; deepseek/kimi/kilo/novita/moonshot/grok/anthropic_api = `false`). Both the Preferences picker and the "Trocar vendor" submenu filter on `vendorEnabled`.

**3. Antigravity reachability.** Rather than ship a broken reachability check, Antigravity is **dropped from the macOS app** in this PR — the binary only discovers its local server on Linux (via `/proc`), so on macOS there's no reachable quota source. Safe macOS discovery (e.g. via `lsof`) is left as dedicated future work; I did not want to mark an unusable setup as configured.

**4. Ring pace arc.** `drawArc` was refactored to `drawArc(fromFraction, toFraction, color)` and draws `[0, min(p,e)]` (calm) + `[min(p,e), p]` (overshoot), so the overshoot continues from the elapsed marker instead of restarting at 12h. The geometry is extracted into a pure `arcAngles(from:to:)` helper.

**5. Docs.** `README.md`, `macos/README.md`, `macos/INSTALL.md`, and `CHANGELOG.md` now reflect the eleven-vendor scope, balance-only vs. rate-limit rendering, the ring/quick-switch behavior, and the Antigravity macOS limitation.

**6. Test coverage + CI.** Added a pure-logic test harness (`macos/run-tests.sh`) — the single-file app has no Xcode project, so the app is compiled with `-D SWIFT_TEST_HARNESS` (which strips `app.run()` via `@main`) alongside a test file that calls its helpers directly. Coverage: arc geometry (`p=80/e=50`, `p=30/e=50`, zero-length), TOML parsing (bare false/true, inline comment, quoted, custom `api_key_env`, omitted, section scoping), Rust defaults per vendor, and parser balance dispatch per vendor (incl. the Moonshot `km_balance` collision case and Anthropic API with/without limit). The CI macOS job now runs the harness alongside the compile.

Local gate is green: `swiftc -O -parse-as-library`, the test harness (44 assertions), `cargo fmt --check`, `cargo test`, and the GNOME marker test all pass. I did not touch `src/` or `gnome-extension/`, so the Linux/Windows jobs should be unaffected.

@akitaonrails — ready for a re-audit of the updated head ([8e27142](https://github.com/akitaonrails/ai-usagebar/pull/35/commits/8e27142df78ada410b9b88a87664597f3c83c5e3)) when you have a moment.

akitaonrails:
Thanks @lucasliet — real per-vendor balances instead of fake 0% rows, the TOML defaults matching src/config.rs, and the first Swift test harness made this an easy merge. Landed as 5b2bffb.

Two maintainer commits went onto the branch before merging (post-audit): a TOML parser fix (`#` inside quoted api keys truncated the value), doc accuracy passes, and a merge-resolution of the conflict with #36's weekly-only OpenAI windows (Snapshot now carries both `creditBalance` and optional session/weekly). CI green on all three platforms including the new macOS harness.

PR #36, opened 2026-07-24
fix(openai): handle weekly-only usage windows

MarlonPassos-git:
## What

Make OpenAI Codex usage windows data-driven so ai-usagebar handles both the
usual 5-hour + 7-day response and the temporary weekly-only response without a
config toggle.

- identify known 5-hour and 7-day windows from `limit_window_seconds`,
  independent of `primary_window` / `secondary_window` position
- keep absent windows absent instead of materializing a confident 0% window
- automatically use weekly values in the default widget format when the
  5-hour window is absent
- omit absent rows from the tooltip, TUI, GNOME extension, and macOS menu bar
- validate semantic duplicates before caching a live response
- preserve the existing layout and placeholder values when both windows are
  present

## Root cause

OpenAI's public Codex usage documentation still describes a 5-hour rolling
limit and possible weekly limits:
https://learn.chatgpt.com/docs/pricing#usage-limits

During the July 2026 experiment tracked in
https://github.com/openai/codex/issues/32707, the usage response changed from:

```text
primary_window   = 18,000 seconds (5h)
secondary_window = 604,800 seconds (7d)
```

to a weekly-only shape:

```text
primary_window   = 604,800 seconds (7d)
secondary_window = null
```

A sanitized live response captured on 2026-07-23 reproduces that exact shape.
ai-usagebar previously treated wire position as semantic: `primary_window` was
always labeled "Codex 5h", while a missing `secondary_window` became a
fabricated 0% weekly gauge. The related request to make weekly-only behavior
permanent is being discussed in
https://github.com/openai/codex/issues/34035.

## Behavior and compatibility

The automatic classifier avoids a setting users would need to toggle again if
OpenAI ends or expands the experiment. Exact 18,000- and 604,800-second windows
take their semantic slots even if reordered; unknown durations retain the
historical positional fallback, and duplicate semantic windows fail as schema
drift instead of silently overwriting one another.

For weekly-only accounts, `{oai_session_*}` / generic session placeholders are
empty and `{oai_weekly_*}` / generic weekly placeholders contain the real
weekly values. Existing dual-window accounts keep their current values and
default 5-hour headline. Other vendors keep their existing snapshot and
renderer logic; the desktop protocol now treats an empty percentage as an
unavailable row rather than 0%.

### Widget and tooltip

Weekly-only OpenAI widget and tooltip

### TUI

Weekly-only OpenAI TUI

## Testing

- `cargo test --all-targets --locked` — 607 passed, 5 live tests ignored
- `cargo test --test live openai_live -- --ignored --nocapture` — passed
  against the current weekly-only response
- `cargo machete`
- `node gnome-extension/marker-logic.test.mjs`
- live widget check: default output selected the weekly percentage/reset and
  the tooltip contained no 5-hour row

`swiftc` is unavailable on the Linux development host; the existing macOS CI
job compiles `macos/ai-usagebar-menubar.swift`.


akitaonrails:
Thanks @MarlonPassos-git — the duration-based classification is exactly the right call over a config toggle, and the pre-cache duplicate validation made this land fast. Merged as 628671b.

I pushed one small follow-up commit onto the branch before merging (post-audit): registered the cross-vendor `{session_elapsed}`/`{weekly_elapsed}` aliases (macOS pace markers never rendered for OpenAI), named the window-duration constants, and covered the no-windows rendering paths in tests.

PR #37, opened 2026-07-25
fix(tui): redraw on terminal resize (maximize/restore)

maickelc:
## Symptom

Resizing the terminal (fullscreen → restore → maximize again) leaves the TUI painted in a corner of the alternate screen. Pressing `R` redraws correctly; maximize alone does not.

## Root cause

The dedicated crossterm reader thread only forwarded `Event::Key`. `Event::Resize` was discarded (`Ok(_) => {}`), so the `select!` loop never woke and `terminal.draw` never ran after a size change. This gap dates from the reader-thread refactor in #26 — even before that, the poll path only accepted `Event::Key`.

## Fix

- Forward `Event::Resize { cols, rows }` over the same channel as keys.
- On resize, call `terminal.resize(Rect::…)` so Fullscreen clears the viewport and resets the diff buffer; the next loop iteration draws into the new area.
- Prefer `resize()` over `clear()`: `clear()` snapshots the cursor via DSR (`\x1b[6n`) and can hang/fail when the terminal does not answer.

Surgical change only (`+33/−5` in `src/bin/ai-usagebar-tui.rs`); key-handling paths untouched.

## Test plan

- [x] A/B: build without the patch → maximize/restore sticks in the corner until `R`
- [x] Build with the patch → maximize/restore redraws without `R`
- [x] PTY smoke: grow 80×24→160×40 and shrink back; process stays alive, redraw bytes emitted, clean quit
- [x] `cargo test` / `cargo clippy --all-targets -- -D warnings` (local gate green)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,13 @@ Each release is also published at
 
 ## [Unreleased]
 
+### Fixed
+
+- **TUI redraws on terminal resize.** The crossterm reader thread discarded
+  `Event::Resize`, so maximizing or restoring the terminal left the UI painted
+  in a corner of the alternate screen until a manual `R`. Resize events are now
+  forwarded to the main loop, which resizes the viewport and redraws.
+
 ## [0.17.1] — 2026-07-24
 
 ### Added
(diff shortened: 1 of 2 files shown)


akitaonrails:
Thanks @maickelc — clean root-cause fix and a great write-up (the DSR hang rationale for `resize()` over `clear()` is exactly the kind of detail that makes review fast). I pushed a tiny `cargo fmt` + CHANGELOG commit on top and merged as d4dab73. Shipping in v0.17.2.

PR #38, opened 2026-07-26
feat: add Cursor vendor (usage-summary two-pool model)

ohmaseclaro:
## Screenshot

Screenshot 2026-07-26 at 15 46 08

image

*`ai-usagebar-tui` showing the Cursor tab live — **Cursor Ultra**, with the **Cursor Models** (Auto + Composer) and **Other Models** (named / API) pools and the billing-cycle reset.*

---

## What

Adds a **Cursor** vendor that shows this billing cycle's two included-usage pools — **Cursor Models** (Auto + Composer) and **Other Models** (named / API) — as percentages, matching the two bars on the Cursor dashboard:

```
❯ 98·100%          # bar (colored by the worst pool)

  Cursor Ultra
  ──────────────
    Cursor Models   98% used   (Auto + Composer)
    Other Models   100% used   (Named / API models · on-demand off)
    Resets 8d 7h
```

Also surfaces the plan (`membershipType`), the billing-cycle reset, whether on-demand spend is enabled, and unlimited plans.

## How it authenticates (no API key)

Cursor exposes no documented usage API or personal API key, so — like the community extensions (`Dwtexe/cursor-stats` et al.) — this reads the session token the Cursor IDE already wrote to its local `state.vscdb` (SQLite, key `cursorAuth/accessToken`), **read-only**. The JWT's `sub` claim yields the user id; combined with the raw token it forms the `WorkosCursorSessionToken` cookie the endpoint expects. No login step of the user's own — signing into the Cursor IDE once is enough.

Endpoint: `GET cursor.com/api/usage-summary` (undocumented; the one the dashboard's own frontend calls). Verified live against an Ultra account; a `#[ignore]`d smoke test (`cursor_live`) exercises it under `make smoke`.

## Wiring

Opt-in (`[cursor] enabled = true`, defaults off like the other reverse-engineered vendors). Follows the existing vendor module shape (`types`/`fetch`/`vendor` + a small `db` for the token read) and the shared cache/flock/staleness machinery. Wired into:

- the Waybar widget (`--vendor cursor`) and scroll-cycling,
- the TUI panel (a gauge per pool),
- the **macOS menu bar app** (the two pools relabel the session/weekly bars as "Cursor Models" / "Other Models"; other vendors keep their time-window labels),
- `config.example.toml`, README (auth table, vendor matrix, placeholders, endpoint-stability note), and `CHANGELOG.md`.

New placeholders: `{cursor_auto_pct}`, `{cursor_api_pct}`, `{cursor_total_pct}`, `{cursor_plan}`, `{cursor_reset}`, `{cursor_on_demand}`, `{cursor_unlimited}`; generic aliases `{session_pct}` = Cursor Models, `{weekly_pct}` = Other Models.

## Dependency

Adds `rusqlite` with the **bundled** feature (statically links SQLite via `cc`), so no system `libsqlite3` is required and the AUR source build stays hermetic.

## Tests

`cargo test` green (642), `cargo fmt --check` clean, `cargo clippy --all-targets -D warnings` clean for the new code, and the macOS Swift harness (`macos/run-tests.sh`) passes with new Cursor cases. Unit tests cover the SQLite token read, JWT/cookie derivation, `usage-summary` parsing (including over-100% pools, unlimited plans, and schema-drift guards), the cache round-trip, and rendering.

## Team accounts (best-effort)

Team / token-based enterprise accounts report **no** `individualUsage.plan` — Cursor doesn't expose the per-seat request limit on this endpoint for them. Their only percentage source is the prose `autoModelSelectedDisplayMessage` / `namedModelSelectedDisplayMessage` strings the payload also carries (e.g. *"You've used 42% of your included total usage"*), which map to the same two pools. `to_snapshot` now parses those when `plan` is absent, accepting a team snapshot only if **both** messages parse (a single stray message stays schema drift, never a fabricated 0%), and labels it `"<Plan> (team)"`. **Unverified against a live team account** — inferred from an independent reverse-engineering of the same endpoint; a payload that matches neither shape still falls through to the existing schema error. Also fixes a latent bug found along the way: `individualUsage.onDemand` lacked its serde rename, so `on_demand_enabled` was silently always `false`.

## Not included

- **GNOME extension** — not wired yet (CLI/TUI/macOS only for now).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,14 @@ Each release is also published at
   menu bar app** (its two pools relabel the session/weekly bars as "Cursor
   Models" / "Other Models"), and the config-example/README docs. Adds a
   `rusqlite` (bundled) dependency. Not wired into the GNOME extension yet.
+  **Team accounts** (`membershipType` with no `individualUsage.plan`) are now
+  parsed too, via the auto/named "You've used N% of your included … usage"
+  display-message strings the payload also carries — the only percentage
+  source Cursor exposes for those accounts, per an independent
+  reverse-engineering of the same endpoint. Unverified against a live team
+  account (labeled `"<Plan> (team)"` in the UI so it's visibly a best-effort
+  path); falls back to the existing schema error rather than a fabricated
+  0% if the display messages don't parse.
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,23 @@ Each release is also published at
 
 ## [Unreleased]
 
+### Added
+
+- **Cursor vendor.** Shows this billing cycle's two included-usage pools —
+  **Cursor Models** (Auto + Composer) and **Other Models** (named / API) — as
+  percentages, from `GET cursor.com/api/usage-summary`, the same undocumented
+  endpoint the Cursor dashboard's own frontend calls. Also surfaces the plan
+  (`membershipType`), the billing-cycle reset, whether on-demand spend is on,
+  and unlimited plans. No API key: the session token is read **read-only** from
+  the local `state.vscdb` SQLite database the Cursor IDE already wrote after you
+  signed in there (the JWT's `sub` claim yields the user id; combined with the
+  raw token it forms the `WorkosCursorSessionToken` cookie the endpoint
+  expects). Opt-in (`[cursor] enabled = true`) and wired into the Waybar widget,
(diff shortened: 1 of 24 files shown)


akitaonrails:
Superseded by and integrated through #44, now merged in 97400ea. The Cursor implementation (including team-account fallback) is in main, together with the maintainer hardening added during final audit: account/cycle-bound caches, required totalPercentUsed validation, safe integer narrowing, regression coverage, and green Ubuntu/macOS/Windows CI. Closing this variant branch so there is one canonical implementation.

PR #41, opened 2026-07-26
feat(anthropic): multi-account auto-discovery via accounts_dir

ohmaseclaro:
## What

Adds **`[anthropic] accounts_dir`** — point it at a directory and each immediate
subdirectory containing a `.credentials.json` becomes an Anthropic account,
labeled by the subdirectory name. This is exactly the layout Claude Code's own
`CLAUDE_CONFIG_DIR` produces, so:

```
CLAUDE_CONFIG_DIR=<dir>/<label> claude   # log in once per account
```

makes `<label>` show up as a TUI tab / `--account <label>` with **no config
edit**, each refreshed independently.

## How

- New `accounts_dir` key on `[anthropic]`; each `<label>/.credentials.json`
  under it is discovered at load time and merged with any explicit
  `[[anthropic.accounts]]` (explicit wins on a label clash). A missing or
  unreadable directory is ignored (never fatal).
- Reuses the existing per-account cache + refresh + TUI-tab machinery — no new
  fetch path.
- **General by design:** it keys only on the standard Claude Code directory
  layout, so any tool that manages multiple logins works with it — not tied to
  one specific account switcher.

## Tests / docs

- Hermetic tests (temp dirs only — no real `$HOME`/creds).
- `config.example.toml`, `README.md`, `CHANGELOG.md` updated.

---

Independent of the Cursor / swap-shortcut / overview PRs — branches off `main`,
touches only Anthropic account discovery. (It does share `config.rs` /
`tui/app.rs` / `CHANGELOG.md` regions with the Cursor PR, so whichever merges
second will need a trivial conflict resolution.)

## Update: macOS Keychain-scoped logins now resolve

On macOS, `CLAUDE_CONFIG_DIR=<dir> claude` stores the login in the Keychain under `Claude Code-credentials-<hash>` (first 8 hex chars of sha256 of the config dir's absolute path) and never writes `<dir>/.credentials.json` — so a discovered/named account could look logged-in via `claude` yet error here. Named accounts now resolve **keychain-first**: the item hashed from their own config dir wins, with the file as fallback (the Linux layout). Keychain-first matters because hand-copied credential files share their refresh-token lineage with the original and die with a 401 as soon as the real holder rotates it. Refresh write-backs follow the read source, so each account keeps one source of truth shared with Claude Code. The per-directory hash can never match a different account's item, so #15's cross-account concern doesn't reapply.

Known limitation (follow-up candidate): *discovery* via `accounts_dir` still keys on `<dir>/<label>/.credentials.json` existing, so a keychain-only login dir isn't auto-discovered yet — declare it in `[[anthropic.accounts]]` and resolution finds the Keychain item.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,24 @@ Each release is also published at
 
 ### Fixed
 
+- **Named/`accounts_dir` Anthropic accounts now find macOS Keychain-backed
+  logins.** `CLAUDE_CONFIG_DIR=<accounts_dir>/<label> claude` was documented
+  to make `<label>` "just work", but on macOS Claude Code stores the login in
+  the Keychain (service `Claude Code-credentials-<hash>`, hashed from the
+  config dir's absolute path) and never writes `<label>/.credentials.json` —
+  so a named account could look logged in via `claude` yet ai-usagebar kept
+  reporting "no usable cache" / stale file errors. Named accounts now prefer
+  the Keychain item hashed from their own directory, falling back to the file
+  (the Linux layout). Keychain-first matters: a `.credentials.json` copied by
+  hand shares its refresh-token lineage with the original, and dies with a
+  401 as soon as the real holder rotates it — reading the file first kept
+  resurrecting those dead snapshots over the live login sitting in the
+  Keychain. Token refreshes write back to the same scoped item, so
+  ai-usagebar and Claude Code keep sharing one source of truth per account.
+  A different account's item can never match (the hash is per-directory), so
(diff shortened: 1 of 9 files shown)


akitaonrails:
Superseded by and integrated through #44, now merged in 97400ea. Main contains both the config-dir-scoped macOS Keychain resolution and accounts_dir support. The known limitation noted here was also fixed during final audit: every immediate account directory is discoverable even when Claude Code stores credentials only in Keychain and creates no .credentials.json. Hermetic Rust/Swift coverage and all three CI platforms are green.

PR #46, opened 2026-07-27
feat(account): add `account add` to capture the active Claude account

### LINKED ISSUE #45: Add an `account add` command to capture the active Claude account
## What I'd like

A native command to capture the Claude account currently logged into Claude Code and register it as a named `[[anthropic.accounts]]` entry:

```sh
ai-usagebar account add work
```

The config-driven multi-account support (#14) is great, but *adding* an account is fully manual today: log in, copy `~/.claude/.credentials.json` somewhere stable, then hand-edit `config.toml` to add the `[[anthropic.accounts]]` block. I've been scripting exactly this by hand.

## Why it fits

The building blocks are already in the tree — `anthropic::keychain` reads the active macOS credential, `anthropic::creds` reads the Linux file, and the Settings overlay already edits `config.toml` with `toml_edit`. An `account add <label>` just wires them together: capture the active account → save a labelled creds file → append an idempotent `[[anthropic.accounts]]` entry. It feels like the natural finish to the multi-account feature.

Happy to send a PR.

zanlucathiago:
Closes #45.

Adds `ai-usagebar account add <label>`: capture the account currently logged into Claude Code and register it as a named `[[anthropic.accounts]]` entry, so `--account <label>` and the TUI's per-account tabs pick it up without hand-editing config.toml.

What it does:

- reads the active credentials (login Keychain on macOS via the existing `anthropic::keychain`, `~/.claude/.credentials.json` on Linux via `anthropic::creds`) and refuses a blob with no `refreshToken`;
- saves them to a labelled file in an `accounts/` dir beside `config.toml` (owner-only perms);
- appends an idempotent `[[anthropic.accounts]]` entry with `toml_edit`, preserving existing comments/keys, and re-capturing the same label is a no-op.

The widget is untouched: the subcommand is optional, so a bare `ai-usagebar ...` still runs as the usage widget and every existing flag parses exactly as before.

Tests cover label validation, the refresh-token check, and the config registration (append, idempotency, existing-content preservation) with injected temp paths — no real `$HOME` access, per the hermetic-test rule. `cargo test`, `cargo clippy`, and `cargo fmt --check` are green.

### DIFF
--- a/src/account.rs
+++ b/src/account.rs
@@ -1,0 +1,254 @@
+//! `ai-usagebar account add <label>` — capture the Claude account currently
+//! logged into Claude Code and register it as a named `[[anthropic.accounts]]`
+//! entry, so `--account <label>` and the TUI's per-account tabs pick it up with
+//! no hand-editing of config.toml.
+//!
+//! Claude Code keeps only one account live at a time (the login Keychain on
+//! macOS, `~/.claude/.credentials.json` on Linux). To watch several
+//! subscriptions at once you log into each with `claude` → /login and capture
+//! it while it is active; ai-usagebar then refreshes that file's token on its
+//! own. This is the native version of the capture people were scripting by hand
+//! around #14.
+
+use std::path::{Path, PathBuf};
+
+use toml_edit::{DocumentMut, Item, Table, value};
+
+use crate::anthropic::creds;
+use crate::cache::atomic_write;
+use crate::config;
+use crate::error::{AppError, Result};
+use crate::widget::cli::AccountAction;
+
+/// Dispatch `ai-usagebar account …` and map the outcome to a process exit code.
+/// This is a real subcommand, not the Waybar widget, so a failure exits non-zero
+/// (the widget's always-exit-0 rule does not apply here).
+#[must_use]
+pub fn run(action: &AccountAction) -> i32 {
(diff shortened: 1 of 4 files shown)


akitaonrails:
Thank you for implementing the requested `account add` command shape and the idempotent TOML update. After auditing it against the current credential model, I’m closing this in favor of #48, which is now merged.

The blocker in this implementation is copying the currently active OAuth refresh token into a second file. Claude/ai-usagebar can rotate that refresh-token lineage, so two independently refreshed copies can invalidate or strand one another; on macOS it also bypasses the named account’s config-dir-scoped Keychain source. #48 keeps the useful `ai-usagebar account add <label>` UX, but launches `claude` with an isolated `CLAUDE_CONFIG_DIR`, so credentials are created at the named account’s canonical source and the default login is never copied or touched.

I retained the subcommand-oriented UX requested here, and #45 is being closed as completed by that safer flow.

zanlucathiago:
Makes total sense. I hadn't fully thought through the refresh-token rotation problem, and in hindsight copying it into a second file was asking for trouble. The isolated `CLAUDE_CONFIG_DIR` approach in #48 is a much cleaner fix, and I'm glad the `account add` UX made it through. Thanks for the careful review and the clear write-up.

PR #47, opened 2026-07-27
Hot-reload config.toml without restarting (menu bar + TUI)

ohmaseclaro:
## What

The macOS menu-bar app and the TUI now **watch `config.toml` and reload live** — enable a vendor, add an `[[anthropic.accounts]]` entry, or tweak an `[ui]` knob and it takes effect within a second or two, instead of requiring an app restart / TUI relaunch after every edit.

- **macOS menu bar** — a native `DispatchSource` file-system watch on the resolved config path (instant). It **re-arms across an editor's atomic save** (write-temp + rename unlinks the watched inode, so a `.delete`/`.rename` event triggers re-opening the new file), debounces the vnode-event burst (~0.3s), then rebuilds the *Trocar vendor* submenu, the ⌥⌘\ swap ring, and the Overview, and queues a refresh.
- **TUI** — a 2s mtime poll added to the existing `tokio::select!` loop, reusing the same reload primitive the Settings-save already uses (`Config::load` → `set_tabs(tabs_from_config)` → `spawn_all`). The file-watch reload **preserves the current tab** (clamped); only an explicit Settings save reselects the configured primary.
- **Safe mid-edit** — a half-written / unparseable file is ignored; the running config is kept until the file parses again, so a broken save never bounces you back to defaults.

## Why not the `notify` crate for the TUI?

`config.toml` changes a handful of times a session. An mtime poll on the loop's existing timer is a few lines and **no new dependency**; the menu bar already covers the instant case natively via Foundation.

## Verification

- `cargo test` — 665 passing; `cargo clippy --all-targets --locked -- -D warnings` clean.
- Swift harness (`./macos/run-tests.sh`) green.
- Live-tested the hardest path: after an **atomic replace** of `config.toml` the app's event-only handle moved to the new inode (fd re-armed) and the app stayed up — proving editor saves keep working, not just in-place appends.

## Notes

Rebased onto `main` after v0.18.0 shipped. `feat/add-custom-claude-account` (a `--add-custom-claude-account <label>` CLI helper that registers a new Claude account in `config.toml`) stacks on top of this branch and pairs with it — adding an account then shows up live.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,9 +18,8 @@ Each release is also published at
   submenu, swap ring, Overview, and TUI tab set all rebuild in place. The menu
   bar watches natively (`DispatchSource`, re-arming across an editor's atomic
   save) so it's instant; the TUI polls the file's mtime every 2s (no new
-  dependency). A half-written/broken file mid-edit is ignored — the running
-  config is kept until the file parses again — so you never get bounced back to
-  defaults.
+  dependency). In the TUI, a half-written/broken file mid-edit is ignored and
+  retried until it parses, so the running config is not replaced with defaults.
 
 ## [0.18.0] — 2026-07-27
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,19 @@ Each release is also published at
 
 ## [Unreleased]
 
+### Added
+
+- **Live `config.toml` reload — no more restart after editing it.** Both the
+  **macOS menu-bar app** and the **TUI** now watch `config.toml` and pick up
+  changes on the fly: enable a vendor, add an `[[anthropic.accounts]]` entry,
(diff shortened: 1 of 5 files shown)


akitaonrails:
I audited this against current `main` and added an integration hardening commit before merge. It now:

- preserves the selected TUI tab by identity when accounts/vendors are inserted ahead of it;
- tracks config path + mtime + size, so canonical/legacy path changes and coarse-mtime rewrites are detected;
- only consumes a file stamp after a successful parse, so a half-written file is retried instead of silently ignored forever;
- narrows the malformed-file changelog guarantee to the TUI, where the full Rust TOML parser enforces it.

Local verification passed: `cargo test --all-targets` (667 unit tests + integration tests), strict Clippy, formatting, and GNOME marker tests.

PR #48, opened 2026-07-27
Add --add-custom-claude-account <label> to register a Claude account

ohmaseclaro:
## What

`--add-custom-claude-account <label>` takes a new custom Claude (Anthropic) account **from nothing to signed-in in one command**:

```bash
ai-usagebar --add-custom-claude-account work
```

1. Appends an `[[anthropic.accounts]]` block to `config.toml` via `toml_edit` — **preserves comments, formatting, other sections**; creates the file if missing.
2. Creates the account's credentials directory.
3. **Launches `claude` to sign in** with that account's own `CLAUDE_CONFIG_DIR`, so the login writes exactly where ai-usagebar reads it back — the config-dir-scoped **Keychain** item on macOS (v0.18.0's keychain-first named resolution), a `.credentials.json` on Linux — **no hand-copying / `security` capture**, and **your default Claude login is never touched**.
4. On return, **re-stamps `config.toml`** so a running menu bar / TUI re-fetches and the account shows up **with data immediately** (via the live-reload from #47).

- **Idempotent** — re-run to sign an already-registered account back in (skips the append instead of erroring).
- `--no-login` registers only (headless boxes, or add-now-sign-in-later).
- If `claude` isn't on `PATH` or the login is cancelled, the entry is still registered and it prints the exact `CLAUDE_CONFIG_DIR=… claude` command to finish by hand. Never touches the default account.

Pure logic (`add_anthropic_account_to_doc`, `doc_has_anthropic_account`, `tildify`, `default_account_credentials_path`, `manual_login_hint`) lives in `config.rs`/`run.rs` with hermetic tests; the login is a thin `Command::new("claude").env("CLAUDE_CONFIG_DIR", dir).status()` that inherits the terminal.

## Verification

- `cargo test` — 673 passing; `cargo clippy --all-targets --locked -- -D warnings` clean; `cargo fmt --check` clean.
- End-to-end against a throwaway `$HOME` with a fake `claude` on `PATH`: register→login success (re-stamps, "signed in"), idempotent re-run, `--no-login` (register + hint), login-exits-1 (registered + retry hint, exit 1), and `claude`-not-on-PATH (registered + hint, exit 1) all behave correctly; spawns with the right per-account `CLAUDE_CONFIG_DIR`.

## Stacking

Stacks on **#47** (live `config.toml` reload) and pairs with it. Both rebased onto `main` after v0.18.0. Merge #47 first.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,16 +11,22 @@ Each release is also published at
 
 ### Added
 
-- **`--add-custom-claude-account <label>`** registers a new custom Claude
-  (Anthropic) account without hand-editing config: it appends an
+- **`--add-custom-claude-account <label>`** takes a new custom Claude (Anthropic)
+  account from nothing to signed-in in one command: it appends an
   `[[anthropic.accounts]]` block to `config.toml` (creating the file if needed,
   preserving comments and formatting via `toml_edit`), creates the account's
-  credentials directory under `accounts/<label>/`, and prints how to sign it in
-  — the macOS Keychain-capture flow, or a `CLAUDE_CONFIG_DIR=… claude` login on
-  Linux. It validates the label and refuses a duplicate before writing anything,
-  and it never touches the default (unnamed) Claude account. Paired with live
-  reload (below), the new account shows up in the menu bar / TUI as soon as its
-  credentials exist — no restart.
+  credentials directory, and then **launches `claude` to sign in** with that
+  account's own `CLAUDE_CONFIG_DIR` — so the login writes exactly where
+  ai-usagebar reads it back (the config-dir-scoped Keychain item on macOS, a
+  `.credentials.json` on Linux) and **your default Claude login is never
+  touched**. When it returns, it re-stamps `config.toml` so the running menu bar
+  / TUI re-fetches and the account shows up **with data immediately** — no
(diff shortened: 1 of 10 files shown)


akitaonrails:
I re-audited this after #47 and pushed a hardening/refactor commit before merge. Key changes:

- exposes the issue-requested `ai-usagebar account add <label>` subcommand (with `--no-login`) and rejects mixed widget/admin flags;
- keeps the safe isolated `CLAUDE_CONFIG_DIR` login instead of copying the active refresh token;
- moves administration out of the Waybar render path and exits non-zero on admin failures;
- quotes manual login commands correctly on macOS paths with spaces and emits PowerShell syntax on Windows;
- creates account directories with mode 0700 on Unix and rejects control/reserved cache labels;
- reuses an existing explicit or discovered account’s real credential path, rather than logging into an unrelated default directory;
- refuses automatic scoped login for legacy custom filenames that Claude cannot populate on Linux;
- fixes a config-loss race: post-login refresh now touches file metadata instead of rewriting the pre-login TOML snapshot over edits made during login.

Verification: 680 Rust unit tests + integration tests passed, strict Clippy and formatting passed, GNOME desktop logic passed, and `cargo machete` is clean.

PR #49, opened 2026-07-27
Per-provider on/off toggle in the Overview (macOS menu bar)

ohmaseclaro:
## What

Each row in the **Overview dropdown** is now a checkbox that controls whether that provider shows in the always-visible top-bar summary.

- **Checkmark = shown; unchecked + dimmed (greyed row) = hidden** from the top bar.
- Hidden providers **stay listed** in the dropdown, so you can turn them back on with a click.
- Hiding some providers also shrinks the top-bar count, so it can draw **mini bars again** instead of falling back to compact %-text.
- The choice **persists** (UserDefaults `overviewHiddenProviders`).
- Because the row click now toggles visibility, **jump-to-a-vendor's-detail** moves to the existing *Trocar vendor* submenu and **⌥⌘\\**.

overview toggles

## Implementation

- Pure `overviewVisibleIds(ids, hidden)` filter (harness-tested); `overviewBarTitle` drops hidden ids before building the summary.
- `renderOverview` sets each row's `.state` (checkmark), greys the row's attributed title when hidden, and points the row action at `toggleOverviewProvider(_:)`, which flips the id in the persisted set and re-renders from the last fetch (no re-fetch — pure visibility).

## Verification

- `./macos/build.sh` clean; `./macos/run-tests.sh` green (added `testOverviewProviderToggle`: nothing-hidden, one-hidden order-preserved, all-hidden, stale-id).

## Notes

Menu-bar-only, independent feature. Branches off `feat/config-watch` (#47) since both touch the Overview menu code; merge #47 first. (The design — checkmark row-toggle vs. an inline switch widget that preserves click-to-jump — was chosen deliberately for the native, low-risk option.)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,15 @@ Each release is also published at
 
 ### Added
 
+- **Per-provider on/off toggle in the Overview (macOS menu bar).** Each row in
+  the Overview dropdown is now a checkbox: click it to drop that provider from
+  the always-visible top-bar summary (checkmark = shown; unchecked + dimmed =
+  hidden). Hidden providers stay listed in the dropdown so you can turn them
+  back on, and dropping some also frees up the top bar to draw mini bars again
+  instead of compact text. The choice persists (UserDefaults). Jumping to a
+  provider's detail view moves to the *Trocar vendor* submenu / **⌥⌘\\** (the
+  Overview row click now toggles instead).
+
 - **Live `config.toml` reload — no more restart after editing it.** Both the
   **macOS menu-bar app** and the **TUI** now watch `config.toml` and pick up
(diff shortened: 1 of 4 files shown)


akitaonrails:
I re-audited this on top of the already-merged #47/#48 work and pushed a small integration hardening commit. The feature is worth keeping, with two fixes:

- the production status-bar filter now goes through the same `overviewVisibleIds` helper covered by the Swift harness, so the test exercises the real selection rule instead of a parallel implementation;
- toggling a visibility checkbox is now truly presentation-only: its UserDefaults notification repaints from the cached Overview without restarting the timer or spawning a fresh provider-wide API refresh.

Local verification passed: 680 Rust tests + integrations, strict Clippy, formatting, GNOME marker tests, and dependency audit. The refreshed macOS CI will compile the AppKit change and run the Swift harness before merge.

PR #65, opened 2026-07-31
TUI: preserve snapshots while refreshing

### LINKED ISSUE #64: TUI: preserve existing usage while auto-refreshing
## Problem

When the TUI auto-refreshes, tabs with an existing snapshot are switched to `Loading` before the request completes. In the Overview this replaces the row with `fetching…`, producing a visible flicker every refresh cycle. It is most noticeable for providers whose request takes longer, such as Claude and OpenAI.

## Question

Would you be open to stale-while-revalidate behavior for the TUI?

That would keep the last successful snapshot visible while the refresh is in flight, use `Loading` only before the first successful result, and update the row when the new result arrives. A subtle refresh or stale indicator could retain visibility into the request state without clearing the data.

I am asking before preparing a PR because this changes the refresh UX and error-state semantics, rather than being only a visual adjustment.

marcelomogami:
Closes #64

## Summary

- keep an existing snapshot visible while its refresh is in flight
- show a static `↻` after `last refresh` while one or more requests run
- retain the last snapshot and mark it stale (`⏸`) if a refresh fails
- keep the initial load behavior unchanged

## Tests

- `cargo test`

### DIFF
--- a/src/bin/ai-usagebar-tui.rs
+++ b/src/bin/ai-usagebar-tui.rs
@@ -450,9 +450,7 @@ fn spawn_one(
     let client = client.clone();
     let cfg = config.clone();
     let generation = app.tab_generation;
-    if let Some(index) = app.tabs_meta.iter().position(|current| current == &tab) {
-        app.tabs[index] = TabState::Loading;
-    }
+    app.begin_refresh(&tab);
     tokio::spawn(async move {
         if !delay.is_zero() {
             tokio::time::sleep(delay).await;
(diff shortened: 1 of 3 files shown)


akitaonrails:
Thank you for preparing this — it crossed with the maintainer implementation in #66, which is now merged on main and closed #64. Your diagnosis and stale-while-revalidate direction were correct. The merged version keeps refresh state per tab, suppresses duplicate requests, preserves the refresh failure diagnostic on the stale snapshot, and covers generation, identity, config-reset, success, failure, and initial-load behavior. Its exact-head local and Ubuntu/macOS/Windows/Rust 1.88 checks all passed. Closing this as superseded rather than merging a second implementation.

PR #71, opened 2026-08-04
feat(kde): KDE Plasma 6 plasmoid — panel widget with per-instance vendor

4nrry:
## Summary

A native KDE Plasma 6 panel widget. Waybar is Wayland-bar-specific and cannot
dock into a Plasma panel, so this ships as a KPackage plasmoid that shells out
to the same binary and draws with Plasma/Kirigami components.

The plasmoid in a Plasma 6 panel showing `5h 2%` and `7d 44%` with their bars, and its popup open below: a Claude / Claude Max 20x header with refresh and TUI buttons, a provider tab strip reading Claude, Codex, Z.AI and OpenRouter with the last two marked by a warning icon, then USAGE & BALANCE over Session (5h) 2%, Weekly (7d) 44% and Fable (7d) 7% — each a full-width bar with its pace text and a live "Resets in" countdown — and an "Updated just now" footer

Rewritten against #83: this is a thin consumer of `usage --json`, not another
frontend carrying its own copy of the shared logic.

- **No vendor knowledge of its own.** Every entry the binary reports becomes a
  tab, with its `display_name` and its plan or its error. Adding a vendor to the
  CLI needs no change here.
- **Severity comes from the report**, not from thresholds re-implemented in QML,
  so the frontends agree about what counts as critical.
- **It does not touch `gnome-extension/`.** No copy of `marker-logic.js`, no
  byte-identity guard, no edit to a file the extension ships.
- **Layout follows `omarchy/Panel.qml`** so the two native panels read the same:
  header, provider tabs, a tinted surface when something is wrong, rows under one
  heading, and a footer. Only the structure is ported — sizes come from
  `Kirigami.Units` and colours from `Kirigami.Theme`, so the widget follows the
  user's Plasma colour scheme, with the five One Dark colours still available.
- **Cheaper than a per-vendor poll.** One report covers everything, countdowns
  tick locally off `reset_at`, and the interval defaults to 300s like your panel.
  Switching tabs is a client-side re-pick, not a refetch.
- **Per applet instance.** The widget never reads
  `~/.cache/ai-usagebar/active_vendor` — that belongs to `--cycle-next`. Two
  instances track two vendors without colliding.

The fetch is wrapped in `timeout -k 5 60`. The executable data engine hands QML
no handle on the child, so a hung binary would keep running with another spawned
every tick; the QML watchdog can only stop us listening. Nearest equivalent to
the extension's `proc.force_exit()`.

## Testing

`make test`, `clippy -D warnings`, `cargo machete` and `make qml-lint` clean.
The CI addition is one step per job that already runs the GNOME node test.

- `make qml-test` instantiates `UsageBar.qml` offscreen under `qmltestrunner`
  and asserts the fill geometry and that the colour follows the report's
  severity rather than a local threshold.
- `make mjs-probe` runs the module in a real applet host, including the one
  thing Node cannot check: that the single-quoted `timeout -k 5 60 …` form
  survives `KShell::splitArgs(AbortOnMeta)` and executes.
- Run in my own panel across two reboots; the screenshot above is this revision.
- Both `.SRCINFO`s verified with the real `makepkg --printsrcinfo`. The packaging
  change is comments only, recording why the CLI package carries no plasmoid
  optdepends.

CI never lints the QML: `ubuntu-latest` is 24.04, which ships Plasma 5.27 and
has no Plasma 6 QML modules — I checked in a container, and without those
imports `qmllint` cannot resolve `PlasmoidItem` or even `Timer.restart`. Point
me at a runner image with Plasma 6 and I will wire it up.

`main.qml` and the settings page are exercised by hand; the applet host injects
`Plasmoid` at runtime and KDE documents no way to mock it, which is why
everything that can be pure lives in `plasmoid-logic.mjs`.

## Not verified

- **Only Anthropic and Antigravity were exercised against live accounts.** I
  have no other subscriptions. Everything else is fixture-tested.
- Visual only: a live Breeze Light↔Dark switch, a vertical panel, a 24px panel,
  and two instances side by side.

`X-Plasma-API-Minimum-Version` is `6.0`, the value
[the KF6 porting guide instructs](https://develop.kde.org/docs/plasma/widget/porting_kf6/);
every property used is in the
[documented Plasma 6 property table](https://develop.kde.org/docs/plasma/widget/properties/).
Developed against 6.6.5.

## One thing dropped

There is no pace marker. `elapsed` exists only inside the report's
human-readable `detail` string, and parsing prose to place a marker is not a
contract worth depending on. The pace text is rendered under each row, so the
information is kept — restoring the marker properly would mean an `elapsed_pct`
field in the report, which I am happy to open separately if you want it.


4nrry:
Saw #83. This is the same shape you turned down there, so before it sits any longer — three options, already costed. Any of them works for me.

**1. In-tree, rebuilt as a thin `usage --json` consumer.**

| | now | after |
|---|---:|---:|
| `marker-logic.mjs` copy | 220 | **0** |
| `plasmoid-logic.mjs` | 662 | ~300 |
| `configVendors.qml` | 134 | **0** |
| rest | 2428 | ~1670 |
| **total** | **3444** | **~2000** |

Counted, not guessed: 469 of `plasmoid-logic.mjs`'s 662 lines are `--format` parsing, the vendor-status probe and the display-name table — all three already in the report as `severity`, `status`/`error` and `display_name`.

Two things beat the line count:

- **The `fix(gnome)` commit stops existing.** No copy of `marker-logic.js`, no byte-identity guard, no edit to a file the GNOME extension ships.
- **Cheaper at runtime.** `omarchy/Panel.qml` polls at 300s and ticks countdowns locally off `reset_at`: 48 vendor-fetches an hour against the 120 this PR does now.

Lost: the pace marker (`elapsed` only lives inside the `detail` string, and your frontend doesn't draw one) and the Vendors login page (the report carries status, not the login actions).

**2. In-tree as it stands.** Only real difference from #83: Noctalia was freezing v4 for a Luau API, so that code would have been born dead. Plasma 6's KPackage is current, and `6.0` is what [KDE's porting guide](https://develop.kde.org/docs/plasma/widget/porting_kf6/) says to declare. Doesn't answer "another frontend to maintain" — your call.

**3. Standalone repo plus a README link**, like @jacksonsieben in #92. Say the word and I'll move it.

Fine with any of them, including no. I'd just rather stop rebasing — six so far, all in `README.md` except one `Makefile` — against a question I never asked you.

4nrry:
Went ahead and built option 1, so it is a diff rather than an offer. Force-pushed over this branch; the description is rewritten to match.

3647 → **2768 lines**, 6 commits → 2, and `gnome-extension/` is now untouched in *every* commit rather than edited and reverted — that was the point, so I rebuilt the history instead of stacking a refactor on top.

What the report replaced: the `--format` field parsing, the awk vendor-status probe, the display-name table, and the severity thresholds. The copy of `marker-logic.js` went with them.

Two things I could not keep:

- **The pace marker.** `elapsed` lives only inside `detail`, and parsing that to place a marker is not a contract I want to depend on. The pace text still renders under each row, so the information survives. An `elapsed_pct` field would bring the marker back properly — happy to open that separately if it is welcome.
- **The Vendors settings page**, which ported the GNOME login flow. `status` and `error` per entry cover the diagnostic half, shown on the tab strip and next to the ring checkboxes.

Rebased onto v1.0.3, CI green on all four. Still fine with option 3 if you would rather this live outside the repo.

PR #98, opened 2026-08-17
plasma: add a native KDE Plasma 6 applet frontend

Marcelo7889:
Plasma was the one desktop without a frontend: the Waybar widget is
wlroots-only and `org.kde.plasma.commandoutput` is not packaged on every
distribution, so Plasma users had no supported way to keep ai-usagebar on the
panel. This adds one under `plasma/`, as a regular KPackage applet
(`Plasma/Applet`) installed with `kpackagetool6`.

Screenshots (panel and popup) are in `plasma/README.md`, added by this PR.

### What it does

- Panel text comes from `ai-usagebar [--vendor X] [extra args] --json`, so
  `--icon` / `--format` keep working. Pango spans are converted to Qt rich text
  attribute by attribute, colors validated, and any other markup escaped.
- Tooltip and popup are built from `ai-usagebar usage --json`: one card per
  provider, a bar per quota window, a marker at the elapsed position of the
  window, the pacing hints, and a **live countdown** from each metric's absolute
  `reset_at` — an open popup keeps ticking between refreshes.
- `ai-usagebar settings show` is read once, at first start, only to learn the
  default provider.

`usage` and `settings` are subcommands, so those command lines are built without
the global options (clap rejects `--vendor` next to them).

### Design notes for review

- **Thin adapter.** No provider table, no product names, no metric ordering in
  the frontend: `display_name`, labels, `value`, `percent`, `severity` and
  `reset_at` are used exactly as the report produces them.
- **Provider switching is applet-local.** It stores `activeVendor` in the
  widget's own config instead of calling `--cycle-next` / `--cycle-prev`. That
  (a) leaves the scroll-cycle state shared with Waybar/TUI alone, (b) lets the
  applet skip providers with no credentials — otherwise a scroll parks the panel
  on a `⚠` — and (c) tells the tooltip which entry to render. Happy to switch to
  the binary's own cycling if you prefer the shared state.
- **Pacing text is recomposed, not passed through.** `detail` is parsed into
  `{elapsed, points, direction}` and rendered through `i18n` so it can be
  translated (a `pt_BR` catalog ships with the applet); anything unrecognized is
  shown verbatim, and the reset fragment is replaced by the live countdown. If
  you would rather have the Rust string rendered as-is, that is a small change
  to `MetricSection.qml`.
- **Hardening.** Provider-controlled text is bounded and stripped of control
  characters and bidi overrides in `Model.js`, and rendered as
  `Text.PlainText`.

### Tests

`plasma/model.test.mjs` follows `omarchy/model.test.mjs`: `Model.js` holds every
pure transformation and no QML globals, so Node covers the report contract
(normalization, spacer dropping, severity fallback, provider cycling, duration
formatting, detail parsing, the Pango conversion) plus the package contract
(every `plasmoid.configuration.*` key declared in `main.xml`, catalog named
after the applet id). Wired into `make desktop-test`, and added to the release
gate in `CLAUDE.md`.

```
node plasma/model.test.mjs
```

### Also updated

- `README.md`: the integrations table and the "native integrations" bullet.
- `Makefile`: new `plasma-test` target, chained from `desktop-test`.
- `CLAUDE.md`: release gate + "What lives where".

Tested on openSUSE Tumbleweed, Plasma 6 / X11, against ai-usagebar 1.0.3 with
Anthropic + OpenAI configured and Z.AI/OpenRouter unconfigured.


Marcelo7889:
Thanks for the quick review window! Looks like #71 (kde-plasmoid/) landed in v1.1.0 while this was open, covering the same ground (and going further — per-panel-instance vendor state, native Kirigami theming, the mjs-in-QML approach). Closing this one to avoid duplicating that work; no need to carry the conflict. I'll keep iterating on the Plasma-specific ideas here in my own repo instead: https://github.com/Marcelo7889/plasma-ai-usagebar

PR #113, opened 2026-08-20
Encrypted sync of local Claude state to a private GitHub repository

ohmaseclaro:
Adds `ai-usagebar sync` — encrypted backup of local Claude state to a private
GitHub repository, so a second machine can continue where the first stopped.
Subcommands: `status`, `setup`, `push`, `pull`, `prune`, `rekey`.

## Design

- **Argon2id** (m=1 GiB, t=3, p=1) derives a key-encryption key; a random master
  key is wrapped under it. BLAKE3 `derive_key` splits that into chunk, name and
  root subkeys. **XChaCha20-Poly1305** seals every object, with the nonce derived
  from the bytes actually sealed and stored inline.
- **Fixed 256 KiB chunks**, content-addressed with a keyed hash, packed into
  **Release assets**. The snapshot pointer flips through a `sha`-preconditioned
  Contents-API `PUT` — a compare-and-swap, so two machines racing cannot lose
  each other's work.
- **The flip is the only commit point.** A push killed before it leaves the
  previous snapshot byte-identical, and the next run re-uploads nothing.
- Release assets over git objects because asset deletion actually reclaims space
  while `gc` is not user-triggerable. Git LFS was rejected on its 10 GiB/month
  ceiling, which disables the whole account when crossed.

## Safety

- **The tool never creates a repository, and structurally cannot** — it is
  configured without `Administration: write`. It *will* offer to add a first
  commit to an empty one, which is a `Contents` operation, and only after the
  private-repo gate has passed.
- **A credential-bearing bundle is refused unless the repository reads private**,
  checked before the first byte and again before the flip. A repository that
  turns public mid-push is an incident: what the run uploaded is deleted and the
  pointer is never touched.
- **`sync pull` is a dry run by default.** A locally-newer file is skipped and
  named; a locally-newer *credential* needs `--force` **and**
  `--force-credentials`. Everything about to be overwritten is archived first,
  and the command that puts it back is printed at the end.
- **No password recovery**, and **changing the password is not revocation** — a
  holder of the old keyfile can still open the bundle with the old password,
  including data written after the change. Both are in the README and in
  `sync rekey`'s help, with a guard test that fails if the sentence is dropped.

## Credentials, and what actually travels

On macOS the **Claude Code login lives in the login Keychain, not in a file**, so
a file-based bundle would have carried nothing usable. It is captured and
restored through a keystore seam, and a restore that would replace a *different*
live login surfaces that for consent rather than doing it silently.

Claude Desktop's `config-tokenCache` is **safeStorage ciphertext bound to the
machine that wrote it**. Restoring it elsewhere would replace a blob that
decrypts with one that never will, so a restore attempts the decryption and
refuses when it cannot — the backup must not be what destroys a working login.
That costs one sign-in on the second machine; re-encrypting it was considered and
rejected, because it would put a bundle-chosen path component in front of a
Keychain-derived key and need a second copy of the path validation.

## Security review

Three adversarial audits, each run from the threat model rather than the test
suite, each followed by remediation. The findings that mattered were two-machine
bugs that needed no attacker:

- The snapshot **counter was computed before the compare-and-swap and never
  recomputed**, so two machines could publish distinct snapshots at the same
  counter — and the rollback anchor read the collision as "already seen",
  silently dropping one machine's backup.
- The **anchor was documented as the rollback defence while having zero reads**
  on the push path. A laundered rollback then let prune irreversibly delete the
  packs it orphaned, executed by the victim, exit 0.
- A machine that had not rekeyed **re-uploaded the old key wrapper** the rekeying
  machine had destroyed, and each push reset its grace window.
- A filename comparison was **byte-exact while macOS folds case**, so
  `.Credentials.json` in a manifest bypassed the credential confirmation and
  `--force` alone could revert a live OAuth token. Fixed as a class:
  `sync::FixedName` has no `PartialEq<str>`, so a byte-exact comparison no longer
  compiles.

## What a real two-machine run then found

The audits could not have found these; only live data could.

- **The local index cached chunk ids under one master key.** Re-running `setup`
  minted a new one, and the planner then reused addresses nothing would ever seal
  — a push that reported success and a bundle that could not be restored.
- **The manifest described what the planner predicted, not what the packer
  sealed.** The bundle's own contents are transcripts being appended to
  continuously; any file that grew between the two produced a named chunk that
  was never uploaded.
- **`PACK_MAX` bounded the writer's body while the reader compared it against the
  published asset**, which carries a header and trailer on top. A real
  50,391,460-byte pack was refused against 50,331,648 and the bundle was
  unreadable.
- **A connection dropping mid-body was classified as an unexpected HTTP 200** and
  never retried, aborting an 880 MiB push at asset 15 of 20.

Each has a regression test that fails on the pre-fix code.

## Surfaces

The macOS menu bar shows sync state read-only and **delegates every action to a
terminal** rather than reach a restore by passing the three flags that skip the
two human confirmations. The TUI Settings overlay ships no sync action for the
same reason. GNOME, KDE and Omarchy have no sync surface; the CLI is the path
there — a recorded decision, not a silent absence.

## Versioning left to you

No version bump, no CHANGELOG release section, no PKGBUILD or `.SRCINFO` change —
the entry sits under `## [Unreleased]` and `Cargo.toml` stays at 1.3.1. Cutting
the release is the maintainer's call, and the checklist in `CLAUDE.md` ties the
tag, both PKGBUILDs and both `.SRCINFO`s together in one step. Branched from
`main` at v1.3.1, so the diff is only what this branch adds.

## Verification

1800 tests, 0 failures. `cargo clippy --all-targets -- -D warnings` clean,
`cargo fmt --check` clean, `make test` green including the GNOME, KDE and Omarchy
contract suites, `./macos/run-tests.sh` 239 assertions, `cargo machete` clean.
Tests are hermetic — the suite passes with `$HOME` and `$XDG_*` unset, so the AUR
`check()` never touches a user's own files or the network.

Exercised end to end against a real private repository: 2.1 GiB of local state,
880 MiB uploaded across 20 packs, a genuine mid-push connection failure resumed
by re-running (12 packs skipped as already present), and a file deleted locally
and restored byte-identically at mode 0600.

**Two things deliberately not claimed as passing:** `omarchy plugin validate .`
was not run — the CLI has no macOS build and this was developed on a Mac, so it
needs a run on Linux. And two calibration probes (`Range:` on private-repo
release assets; asset `state`/`digest` after a torn upload) stay `#[ignore]`d,
each needing a real private repo and token. Both are documented as unmeasured
rather than assumed, and each has a shipped fallback.


akitaonrails:
Thank you for the depth of work in this proposal. The exact audited head contains thoughtful cryptographic, zeroization, path-traversal, and rollback checks, but it is not safe or cross-platform enough to merge as one change.

The blocking findings are:

- the exact-head Windows job fails eight tests; the implementation hard-codes `/usr/bin/tar`, so backup/restore cannot work on a supported platform;
- keyfile KDF parameters are consumed before AEAD authentication. Memory is allowed up to 4 GiB, time/parallelism do not have comparable application ceilings, and the existing memory-budget helper has no production call sites. A hostile or compromised remote keyfile can therefore force excessive memory or CPU work before authentication;
- restore, join, and keyfile-open paths need the same mandatory, realistic KDF limits and available-memory preflight, including a Windows implementation and tests proving oversized parameters are rejected before allocation/work;
- tar stderr is forwarded without the project's usual terminal-control and sensitive-diagnostic sanitation;
- this adds roughly 51,000 lines covering credentials, transcript backup, cryptography, destructive restore/prune, remote GitHub writes, recovery, and rekeying to a usage monitor. That is too broad to establish security and rollback correctness in one review.

I am closing this revision rather than merging red cross-platform CI and a pre-authentication resource-exhaustion path. If the product should own encrypted sync, please split it into small staged PRs (format and bounded KDF/keyfile core first, then local archive/restore, then remote transport and lifecycle), make archive handling portable, and obtain an independent crypto/security review before enabling credential backup. A dedicated companion tool/repository remains the safer ownership boundary.

PR #116, opened 2026-08-21
Report a missing Grok Build CLI on Linux, not a protocol failure

ohmaseclaro:
`spawn()` reports a missing binary as `ErrorKind::NotFound` on macOS, so the "install it or set [supergrok] grok_binary" message fires there. On Linux the child starts through `posix_spawn`, which cannot fail the parent's call: the exec failure surfaces as the child exiting 127, the friendly arm is never reached, and the user is told the ACP protocol failed instead.

`missing_binary_has_a_clear_non_secret_error` has been failing on every Linux job for this reason — verified against untouched `main` inside `rust:1.88` on linux/amd64. The test was right; the code was macOS-shaped.

### DIFF
--- a/src/supergrok/acp.rs
+++ b/src/supergrok/acp.rs
@@ -66,6 +66,19 @@ async fn fetch_billing_inner(grok_binary: &Path) -> Result<BillingResponse> {
 
     let result = run_protocol(stdout, stdin).await;
 
+    // The same "not installed" that `spawn` reports on macOS arrives on Linux as
+    // a child that exited 127: `posix_spawn` cannot fail the parent's call, so
+    // the exec failure surfaces in the child's status instead. Without this the
+    // friendly message above never fires on Linux and the user is told the
+    // protocol failed, rather than that the binary is missing.
+    let missing = matches!(child.try_wait(), Ok(Some(status)) if status.code() == Some(127));
+    if missing {
+        let _ = child.start_kill();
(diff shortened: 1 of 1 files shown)


akitaonrails:
Thanks for investigating this. I reproduced the exact claim against untouched current `main` in the official `rust:1.88-bookworm` Linux image (digest `sha256:af306cfa71d987911a781c37b59d7d67d934f49684058f96cf72079c3626bfe0`).

The read-only, credential-free reproduction ran:

```text
cargo test supergrok::acp::tests::missing_binary_has_a_clear_non_secret_error --locked
```

The base test passes there, on the maintainer Linux host, and on the hosted Ubuntu job: a missing configured executable is returned by `Command::spawn` as `ErrorKind::NotFound`.

The proposed exit-127 fallback is not safe because 127 is also a legitimate exit status from an existing configured executable. It could turn a real CLI/protocol failure into the misleading “CLI not found” message, and the PR adds no negative test protecting that case.

I’m closing this revision because its claimed base failure is not reproducible and the change would broaden the diagnosis incorrectly. A future revision would need:

- the exact distro/libc/container command where untouched `main` fails;
- a regression test that fails on base in that environment;
- a negative test proving an existing executable that exits 127 is not reported as missing;
- a signal tied specifically to executable resolution rather than the child’s generic exit code.

PR #123, opened 2026-08-24
Encrypted bundle format, and a KDF a keyfile cannot choose the cost of (1/3)

ohmaseclaro:
You closed #113 asking for it split — "format and bounded KDF/keyfile core first, then local archive/restore, then remote transport and lifecycle" — and for the pre-authentication resource-exhaustion path closed before any of it ships. This is the first of those, with that path closed.

**7 source files, ~4,720 lines, roughly half of them tests.** No CLI verb, no network, no credential handling, no local collectors.

| file | what it is |
|---|---|
| `crypto.rs` | Argon2id → KEK → wrapped master key → BLAKE3 `derive_key` subkeys; XChaCha20-Poly1305 sealing |
| `chunk.rs` | fixed 256 KiB chunking, zstd, the power-of-two pad |
| `pack.rs` | content-addressed pack assembly |
| `model.rs` | manifest, root, index — the on-disk objects |
| `anchor.rs` | the rollback high-water mark |
| `passphrase.rs` | the password policy, coupled to the KDF cost |
| `mod.rs` | format versions, `FixedName`, the structural guards |

## The finding you raised

KDF parameters live in cleartext in the keyfile and are consumed **before** the AEAD can say the keyfile was forged. That ordering is inherent — the parameters are what produce the key the tag is checked with — so they have to be bounded on their face.

Only `m` was bounded, at 4 GiB. `t` and `p` had no ceiling at all.

**`t` was the sharper half.** It is a pure linear CPU multiplier that allocates nothing, so no memory bound touches it, and `argon2` caps it at `u32::MAX` — about 1.4 × 10⁹ times the shipped cost. A keyfile could ask that opening it take a century and every check here passed it.

| | bound | against |
|---|---|---|
| `m` | **2 GiB** | RFC 9106 §4's first recommended option, and the most any published recommendation asks for. The old 4 GiB was never a bound — it is a guaranteed OOM on the 4 GB aarch64 class this project ships binaries for, so the only machines it protected had enough memory not to need it. |
| `t` | **16** | RFC 9106 gives 1 and 3; OWASP 1–5; borg 3; Bitwarden 2–10. |
| `p` | **16** | Bitwarden's published maximum. Documented as a **tamper signal, not a cost bound** — the vendored `argon2` has no parallel feature, so `p` multiplies neither memory nor work, and calling it a cost bound would be a promise the code does not keep. |

All three sit in the one function every derivation routes through, before anything is allocated or computed. A structural guard asserts there is exactly one `Argon2` construction in the crate and that the check textually precedes it — the previous version of this bound lived in two callers instead, which is how a third caller reached an infallible `vec![]`.

Worth stating because it is what makes this a real gap rather than a style note: **borg and restic — the closest threat model — do not bound these at all.** age and rage do, both as a multiple of their own write default; Bitwarden publishes explicit `(min, max, default)` triples. This follows the latter two.

### Two things I removed rather than finished

The available-memory preflight is **deleted**. Its macOS arm read `hw.memsize` — total installed memory, a constant, so it could never fail — it had no Windows arm, and it had no production call sites.

It also named a `--kdf-memory` flag that **was never built**. Nine places assumed that flag existed, including a live error telling users to re-run with it and a doc line asserting the refusal protects people. All struck, with a test that fails if any message in `crypto.rs` names a flag the build cannot parse. That is the same class of unverified claim you flagged, found while verifying your finding.

## Dependencies — four are genuinely new

I checked this against your `Cargo.lock` rather than asserting it:

| crate | status at v1.4.0 |
|---|---|
| `argon2` 0.5.3 | **new** — zero occurrences, not even transitive |
| `chacha20poly1305` 0.11 | **new** — zero occurrences |
| `blake3` 1.8 | **new** — zero occurrences |
| `zstd` 0.13 | **new** — zero occurrences |
| `zeroize` 1.9 | name present at 1.8.2; this adds a 1.9 |
| `getrandom` 0.4 | name present at 0.2.17 and 0.3.4; this adds a 0.4 |

All six are pure Rust or vendor their C, so the AUR source build stays hermetic — no system libargon2/libsodium/libzstd, nothing to detect at configure time. `argon2` uses `default-features = false` to drop the PHC-string machinery deliberately: the format stores KDF *parameters* but never a verifier, because storing a hash beside the data would hand an offline attacker a plaintext-equivalent oracle.

## Why there is no CLI verb here

You asked for the format first, and a format PR with no user-facing verb is the honest shape of that. Its justification is the test half of the diff: wrong password, downgraded parameters, substituted chunk, flipped bit, truncated manifest, rolled-back counter — each asserting **zero** plaintext reaches the caller. I would rather say that here than have you find unreachable code and infer it.

## The other two, and where the third goes

- **PR 2 + 3 — local archive and restore.** `sync archive` writes a bundle to a directory, `sync restore --from <dir>` reads it back. No token, no network. ~7k lines each; the 12.6k group is too much for one review, which is your point applied rather than argued with.
- **The third — remote transport, lifecycle, and credential backup — goes to a companion repository, not here.** You said a dedicated tool or repository remains the safer ownership boundary. I agree, and would rather you hear it from me than notice it missing. The seam is a `Source` trait: this crate gets the format and a local directory implementation, and the remote one lives elsewhere as a plug-in.

Credential backup stays behind an independent crypto review before it is enabled anywhere it is offered to other people, as you asked.

## Housekeeping

Based on v1.4.0. Four CI jobs green on this exact head, Windows included. No planning files.

Two structural guards here assert non-vacuity by counting files under `src/sync`, and I lowered those floors to what this PR actually contains, with a comment saying so — they must rise as the next PRs land, and I did not want a silently weakened guard reading as a passing one.


akitaonrails:
I've reverted this from `main` in b359ed5. That decision is about ownership, not about the quality of the work, and I want to be precise about which is which.

## What the audit found in the code

The cryptography is sound. The key hierarchy is textbook — Argon2id to a KEK, an AEAD-unwrapped random master key, BLAKE3 `derive_key` for three separately-contexted subkeys. Binding the KDF parameters into the wrap AAD is the correct defence against parameter tampering. Deriving the chunk nonce from the encrypted bytes rather than from the chunk id is a real hazard you found and closed yourselves, and the reasoning in that doc comment is better than most published guidance. `unframe` bounds every length before it slices or allocates, so no decompression bomb can be built out of a valid header. `cargo audit` was clean across all 403 crates.

I also merged a maintainer commit on top (0885668) closing the last of the pre-authentication path: `derive_kek` was still calling `hash_password_into`, whose `vec![Block::default(); n]` aborts through `handle_alloc_error`. The ceiling bounds what a keyfile may *ask* for, but a permitted 2 GiB still kills the process on the 4 GB aarch64 class we ship binaries for — before the AEAD can say the keyfile was forged. Allocating with `try_reserve_exact` and deriving through `hash_password_into_with_memory` turns that into an error on every platform, and dropping the `alloc` feature it needed also drops `password-hash` from the tree. Worth carrying wherever this lands.

## Why it is reverted anyway

The reason is the one from my close of #113, and staging it did not change it:

> this adds roughly 51,000 lines covering credentials, transcript backup, cryptography, destructive restore/prune, remote GitHub writes, recovery, and rekeying to a usage monitor. That is too broad... A dedicated companion tool/repository remains the safer ownership boundary.

Measured on `main` as merged: 4,804 lines and 15 new dependencies (388 to 403 crates, including four cryptographic ones and `zstd-sys`, which compiles C), with **zero call sites reachable from any CLI verb or UI**. That is exactly the shape I asked for, and it is the problem — the project carries the whole cost and users get nothing until parts 2 and 3 land another ~14,000 lines.

You had already concluded that part 3 belongs in a companion repository. I think that reasoning applies to the format too. Splitting them leaves ai-usagebar holding the cryptography, the dependencies, and the review burden while the feature itself lives elsewhere, which is the worst of both arrangements.

One smaller thing pushed in the same direction. `src/sync/mod.rs` documents sixteen submodules, ten of which do not exist here — including `github` ("the only module here that opens a socket"), `keystore`, and `restore`. `docs/sync-format.md` and `tests/sync_vectors.rs` are referenced but absent, and the second is cited as the justification for `Keyfile`'s fields being public. rustdoc reports fifteen broken links from that one directory. None of it is dangerous, but it is the same unverified-claim class you swept out of `crypto.rs`, and it is what a fragment of a larger system looks like when it is documented as the whole.

## What I would suggest

Stand this up as its own crate. It is a coherent, well-tested format that does not need a usage monitor around it, and it would get the independent crypto review on its own terms rather than as a condition attached to someone else's release. If ai-usagebar ever wants encrypted backup, it can depend on it.

Two pieces of your work stayed in the tree, credited: the argon2 fix above, and the source-walking guard helper from `sync/mod.rs`, now at `src/guard.rs` and used by a note-sanitization guard in `claude_desktop`. The observation behind it — that a guard which enumerates fails open on everything added after it — was worth keeping on its own.

Thank you for the rigour here, and for #122, which is merged and shipping. I am sorry to spend your effort this way; the call should have been made at #113 rather than one PR into the split.

PR #129, opened 2026-08-27
feat(commandcode): add Command Code as a vendor

piyush97:
Adds [Command Code](https://commandcode.ai) as a nineteenth vendor, selectable with `--vendor commandcode` and enabled with `[commandcode]` in config.

## What it shows

Command Code meters spend rather than tokens, so its two rolling windows are priced in dollars. The bar leads with the 5-hour window. The tooltip and TUI carry the absolute amounts and the credit ledger:

```
9% · 2h 09m

╭──────────────────────────────────────────────╮
│                     GOAT                     │
│ ──────────────────────────────────────────── │
│  Session (5h)  9% · $1.23 of $14.00 · 2h 09m │
│  Weekly  15% · $5.24 of $35.00 · 6d 16h      │
│                                              │
│  Credits  $49.28 · $20.72 of $70.00 spent    │
│ ──────────────────────────────────────────── │
│   Updated 22:30                              │
╰──────────────────────────────────────────────╯
```

## Endpoints

The same undocumented `/alpha` routes that the official `commandcode` CLI's own `/usage` command calls:

| Route | Supplies |
|---|---|
| `/alpha/whoami` | `org.id`, which scopes the rest when present |
| `/alpha/billing/credits` | the ledger and both `windowLimits` |
| `/alpha/billing/subscriptions` | `planId` |

Only the ledger is required. A personal account with no org, or a subscription lookup that returns 500, still renders its windows.

## Credentials

There is no key to issue. Like Cursor and Kiro CLI, it reuses what a local harness already holds: `~/.commandcode/auth.json` from the official CLI, then omp's, then pi's. `COMMANDCODE_API_KEY` outranks all three, and `[commandcode] auth_paths` overrides the list.

The token is only ever read. Refreshing it belongs to the CLI that owns the file, and writing back from here would race the harnesses that share it. So an expired token is reported as expired instead of being silently rewritten, and one stale harness never masks a live one further down the list.

## Two shape details the parser pins

Both are easy to get silently wrong, so both have named tests.

Resets arrive as millisecond epochs, not the RFC3339 that every other vendor sends. A naive parser drops the countdown without failing.

`windowLimits` has been observed both at the top level and beside the ledger. Either is accepted, and windows are read by name (`fiveHour`, `weekly`) rather than by position.

The API never reports the plan's monthly allowance, so a small plan-to-credits table supplies it. An unrecognised plan keeps its id and omits the "spent of allowance" line instead of inventing a denominator.

## Notes on the invariants in CLAUDE.md

Tests are hermetic. `creds::resolve_from(env, paths)` and `fetch::Endpoints` are the seams, and `creds::resolve` is the production wrapper. Nothing under `#[test]` reads a real `$HOME`, `$XDG` path, or ambient env var.

Formatting goes through the shared helpers. Money uses `format::usd`, report rows use `SectionBuilder::push_metric` so the absolute reset travels with its row, and the label comes from `VendorId::display_name`. No new name or money table in any frontend.

Cache entries are tied to the endpoint plus a one-way SHA-256 token fingerprint, so changing accounts cannot reuse another account's fresh or stale usage. Stale payloads carry the widget through an outage.

The plan string is sanitized and escaped before it reaches the bar or the tooltip.

There are no new dependencies.

## Tests

`make test` goes from 1120 to 1159, all green, with clippy `-D warnings` and `cargo fmt --check` clean.

The 39 new tests cover the parser, the credential search, the endpoint chain (mockito, including no-org, partial failure, 401, cache reuse, cross-account isolation, stale fallback and schema drift), and the renderer. Fixtures are synthetic, matching the `opencode-go` convention.

`tests/live.rs` gains one `#[ignore]`d `commandcode_live`. It skips with a message when no harness on the machine is signed in, so `make smoke` stays green for everyone else. I verified it against the real API.

Docs: a support-matrix row and stability note in `docs/vendor-endpoints.md`, an auth-table row and a Command Code section in `README.md`, a `[commandcode]` section in `config.example.toml`, and a `[Unreleased]` entry in `CHANGELOG.md`. No published changelog section moved.

I left version numbers, PKGBUILDs and `.SRCINFO`s alone, since those belong to your release checklist rather than to a feature PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,10 +37,11 @@ Each release is also published at
   the 5-hour and weekly rolling spend windows — priced in dollars, as Command
   Code meters spend rather than tokens — plus the plan and the monthly credit
   remaining. Like Cursor and Kiro CLI it needs no key of its own: it reuses the
-  OAuth credential whichever local harness is already signed in
-  (`~/.commandcode/auth.json`, then omp's and pi's), with `COMMANDCODE_API_KEY`
-  as an override. The credential is only ever read — refreshing it belongs to
-  the CLI that owns the file — so an expired token is reported as expired
+  OAuth credential from either the official CLI or pi
+  (`~/.commandcode/auth.json`, then `~/.pi/agent/auth.json`), with
+  `COMMANDCODE_API_KEY` as an override. The credential is only ever read —
+  refreshing it belongs to the CLI that owns the file — so an expired token is
+  reported as expired
   rather than silently rewritten.
 
 ## [1.7.0] — 2026-08-25
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,19 @@ Each release is also published at
 
 ## [Unreleased]
 
+### Added
+
+- **Command Code** (`commandcode.ai`) is supported as a vendor, selectable with
+  `--vendor commandcode` and enabled with `[commandcode]` in config. It shows
+  the 5-hour and weekly rolling spend windows — priced in dollars, as Command
+  Code meters spend rather than tokens — plus the plan and the monthly credit
+  remaining. Like Cursor and Kiro CLI it needs no key of its own: it reuses the
+  OAuth credential from either the official CLI or pi
+  (`~/.commandcode/auth.json`, then `~/.pi/agent/auth.json`), with
+  `COMMANDCODE_API_KEY` as an override. The credential is only ever read —
+  refreshing it belongs to the CLI that owns the file — so an expired token is
+  reported as expired rather than silently rewritten.
+
 ### Fixed
 
 - Z.AI's rows in the native panels — Omarchy Quattro, GNOME, KDE and the TUI —
@@ -95,17 +108,6 @@ Each release is also published at
   value column shows the same `{pct}%` every other vendor shows instead of
   raw request counts; the counts move to the footnote next to the remaining
   figure and the reset countdown.
-- **Command Code** (`commandcode.ai`) is supported as a vendor, selectable with
-  `--vendor commandcode` and enabled with `[commandcode]` in config. It shows
-  the 5-hour and weekly rolling spend windows — priced in dollars, as Command
-  Code meters spend rather than tokens — plus the plan and the monthly credit
-  remaining. Like Cursor and Kiro CLI it needs no key of its own: it reuses the
-  OAuth credential from either the official CLI or pi
-  (`~/.commandcode/auth.json`, then `~/.pi/agent/auth.json`), with
-  `COMMANDCODE_API_KEY` as an override. The credential is only ever read —
-  refreshing it belongs to the CLI that owns the file — so an expired token is
-  reported as expired
-  rather than silently rewritten.
 
 ## [1.7.0] — 2026-08-25
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,19 @@ Each release is also published at
 
 ## [Unreleased]
 
+### Added
+
+- **Command Code** (`commandcode.ai`) is supported as a vendor, selectable with
+  `--vendor commandcode` and enabled with `[commandcode]` in config. It shows
+  the 5-hour and weekly rolling spend windows — priced in dollars, as Command
+  Code meters spend rather than tokens — plus the plan and the monthly credit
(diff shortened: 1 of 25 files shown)


piyush97:
Pushed a follow-up docs commit (`8d4806b`) after running this on a real Omarchy install.

I installed it, opened the settings panel, and my first reaction was "I don't see where to put the Command Code API key", even though the vendor was already working and showing live data. The panel renders one key field per key-authenticated vendor, so a provider absent from that list reads as unfinished rather than as one that needs nothing. It looked like the reason there was no data, when it already had all of it.

No code changed. Command Code still takes no key, the same as Claude, Codex, Cursor and Kiro. The commit only says so in the three places someone actually looks: the Quattro panel's README, the settings-overlay module doc, and the Command Code section of the main README.

One drive-by in that same sentence: the module doc's list of local-product-state vendors was missing Kiro, which has authenticated that way since it landed. Happy to split that out if you would rather keep this PR to Command Code alone.

The gate is still green. `make test` passes 1159, and clippy `-D warnings` and `cargo fmt --check` are clean.

akitaonrails:
Audited and pushed two maintainer commits. The vendor itself is good work — hermetic seams, `format::usd`, `push_metric`, the token fingerprint in the cache key, and the two shape details you pinned with named tests are exactly right.

## What I changed

**`apiKey` is now scoped to `.commandcode/auth.json`** (c10c826). `read_from` chained a bare `object.get("apiKey")` after the provider-keyed lookup and applied it to all three paths. I want to be precise about severity, because I got it wrong first: this is **not** an active leak. I checked the two third-party formats, and neither puts a token at a top-level `apiKey` — pi types its store as `Record<providerId, Credential>`, so its top level is provider ids. The provider-keyed lookup you wrote is the correct and safe one.

But `~/.pi/agent/auth.json` holds the user's Anthropic, OpenAI, OpenRouter and Radius credentials, and the token read from it goes on to `api.commandcode.ai` as a bearer. A key that names no provider shouldn't be read out of a multi-provider keystore whatever shape a future version gives it. Scoped to the official file, where the CLI stores a pasted key and the whole document is Command Code's, that fallback is right. Your provider-keyed lookup is untouched and still applies to all three paths.

**Merge composition with #133** (same commit). #133 landed while this was open and moved `{vendor_short}` behind `VendorId::short_name()`, with a guard rejecting duplicate or non-three-letter codes. This vendor shipped `"cc"`, which that guard rejects, and `short_name()` had no arm for the new variant, so the merge didn't compile. It's `VendorId::CommandCode.short_name()` -> `"cmc"` now, read through the shared accessor rather than as a nineteenth hardcoded literal.

## One question before I merge

`.omp/agent/auth.json` — what is it for?

omp's own porting notes say credentials moved off that file:

| Upstream | Our Fork | Notes |
| `proper-lockfile` + `auth.json` | `agent.db` (bun:sqlite) | Credentials stored exclusively in `agent.db` |

and `commandcode` returns no matches anywhere in `can1357/oh-my-pi`. So as far as I can tell that path either never matches or reads a stale pre-fork file.

`.pi/agent/auth.json` checks out fully, for what it's worth — pi documents it as the credentials file, and `commandcode` does appear in pi's `model-resolver.ts`, so a Command Code credential genuinely can live there under its provider key.

Is the omp path from a version I haven't found, or is it carried over from pi's layout on the assumption the fork kept it? Happy to keep it if it's real — I'd just rather not ship a credential-file read we can't point at. Everything else is ready to merge.

piyush97:
You're right. The OMP path was carried over from pi's layout, and I couldn't substantiate it. Current OMP stores credentials in `agent.db`, and I found no Command Code provider in `can1357/oh-my-pi`.

I removed `~/.omp/agent/auth.json` in 9c42c70 and updated the README, config example, changelog, and config docs. The default search is now limited to the two paths we can point at: `~/.commandcode/auth.json`, then `~/.pi/agent/auth.json`.

I also fixed one adjacent edge in c10c826. `read_from` correctly scoped the unkeyed `apiKey` fallback to Command Code's own file, but `message_for` still considered a top-level `apiKey` in pi when deciding whether to report an expired Command Code login. It could not send that token, but it could report the wrong status. The expiry check now uses the same scope, with a test for an unrelated expired `apiKey` in pi.

Checks are green: `make test` (1,162 tests), `cargo fmt --check`, and clippy with `-D warnings`.

Thanks for catching the unsupported path.

PR #132, opened 2026-08-27
feat(kimi): accept a Kimi For Coding subscription login

miillerb:
Use the Kimi Code CLI's own OAuth session as a Kimi credential, so a subscriber who works through `kimi` has nothing to create or paste. An API key still takes precedence wherever one is set, and existing installs are untouched.

Same endpoint either way — verified live: `GET api.kimi.com/coding/v1/usages` with the CLI's bearer returns `200` and the same payload an API key gets.

This is the shape ai-usagebar already uses for **Cursor, Kiro CLI, SuperGrok and Antigravity**: read the local product's own session instead of asking for a credential the user maintains by hand.

## What this adds

- No API key configured → use the OAuth session kimi-code already stored at `~/.kimi-code/credentials/kimi-code.json`.
- `[kimi] credentials_path` — override for a relocated `KIMI_CODE_HOME`, mirroring `[cursor] db_path` / `[kiro] db_path`.
- `[kimi] region` — `auto` (default) follows kimi-code's own `~/.kimi-code/region` marker; `cn` pins api.kimi.com, `global` pins api.kimi.ai. A token minted by one deployment means nothing to the other.
- The plan label now comes from `/coding/v1/me` (`Allegretto`) instead of the raw `LEVEL_*` enum, with the humanized enum as fallback.
- `fetch_snapshot` keeps its signature; the OAuth path is a new `fetch_snapshot_with_auth`.

## One deliberate departure from the Kiro precedent

CLAUDE.md says rotated credentials go to ai-usagebar's own cache, never back to the CLI's store. This writes them **back into kimi-code's credential file**, because `auth.kimi.com` rotates the refresh token on *every* grant (verified: two consecutive grants, two different refresh tokens). The access token lives 15 minutes, so with a private sidecar a 300 s widget tick would leave the CLI holding a superseded token every few minutes and eventually log the user out of their own CLI. Kiro's store is a live sqlite database and its refresh response makes `refreshToken` optional; neither is true here.

The write matches kimi-code's own semantics (atomic, 0600, unknown fields preserved) and runs under the lock kimi-code itself takes — `src/kimi/lock.rs` speaks `proper-lockfile`'s directory-lock protocol rather than `flock`, which would be invisible to the CLI. The file is re-read *inside* the lock, so a refresh that landed while we queued is used instead of redone.

If writing into another tool's credential store is a line the project would rather not cross, say so and I'll cut this back to the display work alone.

## Tests

Unit tests cover the wire shape and logout tombstone, write-back preserving unknown fields and mode 0600, the refresh buffer, region resolution, a rejected refresh that does not echo the body, lock acquire/release plus stale takeover, and — in `fetch` — a valid token used as-is, an expiring one refreshed with the rotation landing in the CLI's file, and a peer refresh picked up instead of rotating again. `tests/live.rs` adds `kimi_subscription_live`, `#[ignore]`d and skipping when the CLI is not logged in, like `kiro_live`.

All tests inject paths and a clock; none reads a real `$HOME` or an ambient env var. `make test` (1170 passing), `cargo clippy --all-targets -- -D warnings` and `cargo fmt --check` are clean, and the whole path was exercised against a real Kimi For Coding subscription including the expired-token refresh.

---

*Correction: this PR originally claimed a Kimi For Coding subscription issues no API key. That was wrong — one can be created at `kimi.com/code/console`, as @lucasliet pointed out. The claim is removed from the code, README, docs and CHANGELOG.*

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,50 @@ Each release is also published at
 
 ## [Unreleased]
 
+### Added
+
+- Kimi accepts a **Kimi For Coding subscription** as a credential: when no
+  `KIMI_API_KEY` (or inline `api_key`) is set, the vendor uses the OAuth
+  session the Kimi Code CLI already stored at
+  `~/.kimi-code/credentials/kimi-code.json` — so a subscriber who works through
+  the CLI has nothing to create, paste, or rotate by hand. Same
+  `/coding/v1/usages` endpoint, same weekly + 5h windows; an API key still takes
+  precedence when one is set. New optional `[kimi] credentials_path` (for
+  a relocated `KIMI_CODE_HOME`) and `[kimi] region` — `auto` follows
+  kimi-code's own `~/.kimi-code/region` marker, `cn` pins `api.kimi.com`,
+  `global` pins `api.kimi.ai`.
+
+  The CLI's access token lives 15 minutes, so ai-usagebar refreshes it against
+  `auth.kimi.com/api/oauth/token` and writes the rotated pair **back into
+  kimi-code's own credential file** (atomically, mode 0600) rather than into a
+  private sidecar the way the Kiro vendor does. Kimi rotates the refresh token
+  on every grant: a private copy would leave the CLI holding a superseded token
+  after each widget tick and log the user out of their own CLI. The refresh
+  runs under kimi-code's own `proper-lockfile` protocol so the two clients
+  never rotate at once, and the file is re-read inside the lock so a refresh
+  that landed while waiting is used instead of being redone.
+
+- Kimi's plan label now reads the subscription's own tier name ("Andante",
+  "Moderato", "Allegretto", "Allegro") from `/coding/v1/me`, fetched
+  concurrently with `/usages` so a widget tick still costs one round-trip.
+  `/usages` only carries the `LEVEL_*` wire enum, which was what the bar used
+  to show. When `/me` is unreachable or the account has no coding profile, the
+  enum is humanized instead (`LEVEL_INTERMEDIATE` → `Intermediate`) — never
+  mapped to an invented tier name. A still-fresh cache entry holding a raw
+  `LEVEL_*` plan is refreshed once rather than waiting out its TTL. The
+  profile response's personal fields (email, phone, nickname) are never
+  deserialized, so they cannot reach a snapshot, the cache, or an error
+  message.
+
+### Changed
+
+- Kimi's default widget format now shows both independent quotas —
+  `5h X% · 7d Y%` — instead of a single percentage that silently discarded
+  one of them. A custom `format` in config.toml is untouched. The TUI panel's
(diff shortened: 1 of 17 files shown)


lucasliet:
Kimi for Coding actually issues an api key for login..
that`s  the official way to login with it on opencode for example

you can make one here https://www.kimi.com/code/console
on API KEYs section
image

you can call the model like
```bash
 curl -s "https://api.kimi.com/coding/v1/chat/completions" \     
 -X POST \
 -H "Content-Type: application/json" \     
 -H "Authorization: Bearer $KIMI_API_KEY" \ 
 -d '{               
    "model": "k3-256k",                                               
    "messages": [
      {"role": "user", "content": "Answer with \"OK it worked\""}
    ]
 }'
```
which is already supported on ai-usagebar
image

miillerb:
You're right, and thanks for the console pointer — I inferred "no API key" from the CLI using OAuth and never checked the console. The claim had spread to eight places (code docs, README credential table, `docs/configuration.md`, `config.example.toml`, CHANGELOG); e404c05 removes it and the PR description now opens with the correction.

What the PR still buys: it drops the key-management step for people already logged into the `kimi` CLI — the same shape ai-usagebar already supports for Cursor, Kiro CLI, SuperGrok and Antigravity, which all read a local product's own session instead of asking for a credential the user maintains by hand. An API key keeps precedence when one is set, so existing installs are untouched.

The cost is real and worth your scrutiny: `auth.kimi.com` rotates the refresh token on **every** grant, so the vendor writes the rotated pair back into kimi-code's own credential file (atomically, 0600, under its `proper-lockfile` directory-lock protocol) instead of keeping a private sidecar the way the Kiro vendor does. With a sidecar, a 300s widget tick would leave the CLI holding a superseded refresh token every few minutes and eventually log the user out of their own CLI. That is a deliberate departure from the Kiro precedent in CLAUDE.md, and if writing into another tool's credential store is a line this project would rather not cross, then the API-key path you describe is the better answer — in that case this PR should shrink to just the display work (plan name from `/me`, both quotas by default, panel percentages).

One thing you'd know better than me: does a console key report the subscription's own weekly quota, or a separate pay-as-you-go pool?

miillerb:
Fixed, thanks. The "no API key" claim is out of the code, README, `docs/configuration.md`, `config.example.toml` and the CHANGELOG, and those now point at `kimi.com/code/console` as where a subscription can issue one. An API key still takes precedence over the CLI session wherever one is set, so the path you describe is untouched and remains the one used whenever a key is configured.

What the OAuth path covers is the case where no key is configured: it reads the session `kimi` already stored, the same way ai-usagebar reads Cursor's, Kiro CLI's, SuperGrok's and Antigravity's local sessions. The no-credentials error now also names the file it looked in, since `[kimi] credentials_path` and `KIMI_CODE_HOME` both move it.

The PR itself is smaller since your comment: squashed to a single commit, and the description trimmed to the essentials.

PR #142, opened 2026-08-30
feat(plasmoid): add the AI Usage cards widget for KDE Plasma 6

rusagent:
## What

Adds a second KDE Plasma 6 widget under `plasmoid/` alongside the existing `kde-plasmoid/` — submitted as an **alternative** design, fully aware the repo already ships a KDE integration; happy to rework, merge features into `kde-plasmoid`, or drop this if the maintainer prefers one widget.

## Design

Instead of provider tabs over the fixed `usage --json` report, this widget renders **one card per vendor** from the vendor JSON API (`ai-usagebar --json --vendor <v>`), so each card owns its credential state:

- session (5h) and weekly (7d) gauge bars with color-coded severity (green → amber → red at 45/70/90%)
- a window a vendor does not report renders as `—` + *not reported*, never a fabricated 0% bar (e.g. OpenAI's dropped 5h window lights back up on its own)
- distinct card states: **loading / ok / stale** (data older than two refresh cycles is flagged next to the timestamp) / **unconfigured** (muted, with the CLI's actionable message in the hover tooltip) / **error** (red, failure text inline) — so a silently failing fetch cannot pass as current
- per-card refresh on click, toolbar refresh, auto-refresh every 5 minutes, and a refresh on popup expand after resume
- cross-vendor `{session_pct}`/`{weekly_pct}` aliases (registered since 0.17.0) keep the QML to one command template; adding a vendor is one `ListElement`

## Relationship to `kde-plasmoid/`

| | `kde-plasmoid/` | `plasmoid/` (this PR) |
|---|---|---|
| Data | fixed `usage --json` report, client-side vendor selection | per-vendor `--json --vendor`, per-card credential state |
| Layout | provider tabs | stacked dark cards with gauges |
| Config | KConfigXT per applet instance | none — every configured vendor gets a card |

Both shell out to the `ai-usagebar` binary (discovered on `PATH` and the usual install locations) and read no credentials themselves. The plasmoid has been running daily on Plasma 6 as the author's primary widget (v0.3.0 → v0.4.1).

## Out of scope of this PR

The author also runs a local `scripts/auto-update.sh` + systemd user timer/path units that rebuild and `kpackagetool6 --upgrade` the widget automatically; those hardcode a repo location and are not portable enough to ship here, but can be contributed separately if there is interest.

## Validation

- `kpackagetool6 --type Plasma/Applet --install/upgrade` verified on Plasma 6 (install, upgrade-in-place, and plasmashell-restart flows)
- QML is data-driven off the existing vendor JSON contract; no Rust changes

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,6 +40,7 @@ codebase.
 - [Format placeholders](docs/format-placeholders.md)
 - [Provider endpoints and live tests](docs/vendor-endpoints.md)
 - [KDE Plasma 6 plasmoid](kde-plasmoid/README.md)
+- [AI Usage cards plasmoid](plasmoid/README.md)
 
 ## Install
 
@@ -464,6 +465,7 @@ privileges, and does not overwrite user configuration.
 | [macOS menu bar](macos/README.md) | Claude, Codex, Z.AI, OpenRouter, DeepSeek, Kimi, Kilo, Novita, Moonshot, Grok (xAI), Anthropic API, Cursor, Google Antigravity | Thirteen providers. |
 | [GNOME Shell](gnome-extension/README.md) | Claude, Codex, Z.AI, OpenRouter, DeepSeek, Google Antigravity | Antigravity's two quota pools appear as grouped rows. |
 | [KDE Plasma 6](kde-plasmoid/README.md) | Whatever `usage --json` reports | Provider tabs in the popup; vendor is per applet instance. |
+| [AI Usage cards](plasmoid/README.md) | Claude, OpenAI, Z.AI, Kimi, SuperGrok | Alternative dark-card widget: one card per vendor with gauges, per-card refresh on click, auto-refresh every 5 minutes, stale/unconfigured/error card states. |
 
 Cursor is not available in the GNOME extension yet. On GNOME, use
 `ai-usagebar --vendor cursor` or open the TUI.
--- a/plasmoid/README.md
+++ b/plasmoid/README.md
@@ -1,0 +1,84 @@
+# KDE Plasma widget
+
+The widget requires Plasma 6 and an installed `ai-usagebar` binary. It searches
+`PATH`, `~/.local/bin`, `~/.cargo/bin`, `/usr/local/bin`, and `/usr/bin`.
+
+Install or update the widget from the repository root:
+
+```bash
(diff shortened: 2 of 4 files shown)


akitaonrails:
Thank you for this — the card design is genuinely good, and the QML is careful in the places that matter: `shellQuote` handles embedded quotes correctly, the vendor strings come from a fixed model rather than user input, and the unconfigured-vs-error card distinction is a nicer failure mode than what the existing widget does.

I can't merge it as a second plasmoid, though, and I want to be straight about why rather than leave it sitting open.

**The repo already ships a KDE Plasma 6 plasmoid** — `kde-plasmoid/`, plugin id `io.github.akitaonrails.ai-usagebar`. This adds a second one at `plasmoid/` under `dev.altendorfer.aiusage`. Two widgets doing the same job in one repository means every future vendor, every report-schema change, and every packaging fix has to be done twice, and it leaves users with a question I can't answer for them: which one do I install? That cost lands on me permanently, so it's mine to decline.

Two other things would need to change regardless of the packaging question:

1. **`main.qml:180-247` hardcodes a five-vendor table** — label, short name, and window labels for Claude, OpenAI, Z.AI, Kimi and SuperGrok. `CLAUDE.md` is explicit that frontends must not carry a provider-name table: canonical names and metric projection belong in Rust, and `usage --json` already emits `short_name` and `display_name` per entry precisely so a frontend doesn't restate them. As written the widget also silently covers 5 of 19 providers, and gains nothing when a new one lands.

2. **No tests.** `kde-plasmoid/` keeps its pure logic in `plasmoid-logic.mjs` with `plasmoid-logic.test.mjs` wired into `make test`. 692 lines of QML with no coverage isn't something I can maintain.

**What I'd happily take instead:** the card layout as an alternative view inside `kde-plasmoid/`, driven by the vendor list `usage --json` already returns rather than a table, with its formatting logic in the existing `.mjs` adapter so the Node contract tests cover it. That keeps the part of this PR that's actually valuable — the visual design, which is better than what's there — and drops the part that's a second copy of plumbing that already exists.

If you'd rather ship it standalone, please do: publish it to the KDE Store under your own id and I'll link it from the README's frontend table as a third-party widget. That's a real option and not a brush-off — it just shouldn't live in this tree.

Closing for now. Reopen or open a fresh PR if you want to take the rework route, and I'll review it properly.

PR #150, opened 2026-09-05
feat(omarchy): show every provider in the bar with official marks

calebdw:
## What this changes

Hello!

I typically use 3 different providers on different things so it would be nice to see all of them at once in the top bar---so I added a config option for it. Since I was already adding the official provider icons, I went ahead and updated what's shown in the panel.

Thanks!

## Checklist

- [ ] A test that fails on `main` and passes here, for a bug fix

## Testing

Tested in Omarchy Linux:

image

image

image


akitaonrails:
Thanks — and thanks for using the PR template; you're the first, and filling it honestly (leaving the bug-fix test box unchecked because this is a feature) is exactly right.

The Rust side is done the way I'd want it: `bar_icon()` lives on `VendorId`, reaches the frontend through `usage --json`, and no provider table lands in QML. The SVGs are clean — pure path geometry, no scripts, no external refs. `manifest.json` adds a setting key rather than touching the version, which is the release process's to own.

I've pushed two small things (`f9d9b20`): rustfmt over a `report.rs` test, and the changelog bullet moved under `### Added` — it had no category heading, and the release process reads those to pick the next version, so an additive feature with no heading would have been invisible to it.

**One thing I can't merge yet.** `bar_icon()` gives 20 providers 11 distinct glyphs:

```
󰚩  x8   anthropic, zai, kimi, moonshot, minimax, nous, opencode-go, commandcode
󰧑  x2   deepseek, antigravity
󰇷  x2   grok, supergrok
```

Your four SVG marks cover Claude, Codex, Copilot and Grok, so after `BrandMark` resolves them **seven providers still render as the same robot** and DeepSeek/Antigravity share one. In a bar whose whole purpose is showing every provider side by side, a third of them can't be told apart — which is the one thing the feature has to get right.

The existing test reads like it guards this, but asserts three hand-picked pairs that all happen to be in the distinct set, so it passes. Here's one that actually holds the line — it currently fails, listing the collisions:

```rust
/// The bar can show every provider at once, so a glyph two providers share
/// tells the user nothing about which row is which. Grok and SuperGrok are
/// the one sanctioned pair — same brand, two products.
#[test]
fn every_vendor_has_a_bar_icon_and_no_two_share_one() {
    use std::collections::BTreeMap;

    let mut by_icon: BTreeMap<&str, Vec<&str>> = BTreeMap::new();
    for vendor in VendorId::all() {
        assert!(!vendor.bar_icon().is_empty(), "{}", vendor.slug());
        by_icon.entry(vendor.bar_icon()).or_default().push(vendor.slug());
    }

    let shared: Vec<_> = by_icon
        .iter()
        .filter(|(_, vendors)| vendors.len() > 1)
        .filter(|(_, vendors)| vendors.as_slice() != ["grok", "supergrok"])
        .collect();
    assert!(
        shared.is_empty(),
        "these providers are indistinguishable in a bar that shows them \
         side by side: {shared:#?}"
    );
}
```

**I'm handing this back rather than fixing it myself on purpose.** Choosing nine more Nerd Font glyphs blind is a bad trade: a glyph missing from the user's font renders as a tofu box, which is worse than a duplicate robot, and I can't see them render. You can.

If distinct glyphs are hard to find, there's an option that needs no font at all: fall back to the provider's **`short_name`** (`zai`, `kmi`, `msh`, `mmx`, `nrs`, `ocg`, `cmc`). It's already on the entry in `usage --json`, and `vendor.rs:334` guarantees no two providers share one — so it's distinct by construction and can't go tofu. A three-letter chip beside the usage would read fine next to the brand marks.

**One more thing before merge:** where did the four SVGs come from? This is an MIT repo with no prior bundled assets, and they're third-party marks (Anthropic, GitHub, xAI, OpenAI). A one-line provenance note — source and its licence — next to them or in `omarchy/README.md` is all I'm after. I'd rather ask than guess and attribute them wrongly.

Everything else is ready. Ping me when the glyphs are sorted and I'll merge.

akitaonrails:
Rebased onto `main` for you (`c7f4f95`) — #149 landed and the two changelog entries collided. Both sit under `### Added` now, #149's first since it merged first, and the released section is untouched.

Nothing else changed. The icon collision from my review above is still the one thing left here, and it's still yours: seven providers render as the same `󰚩` once the SVG marks resolve, and I can't pick replacement glyphs I'm unable to see render.

calebdw:
@akitaonrails  I've made the updates :+1:

akitaonrails:
This is the fix, and you went further than I asked — thank you. Audited the new head (`65c89e7`):

**The collision is gone.** Every provider is now distinguishable: 18 SVG marks resolve through `BrandMark`, and the three-letter `short_name` chip covers what has no mark. I checked the wiring rather than trusting it — walked all 19 slugs through `Model.brandIconFile()` and confirmed each returned filename exists in `omarchy/icons/`. No dangling reference, so nothing renders blank. Command Code is the only glyph fallback, which matches your README note.

**The assets are clean.** Scanned every SVG for `script`, `onload`, `href`/`xlink`, `foreignObject`, `<image>`, `data:`, `@import`, `<!ENTITY>` — nothing. All regular files, no mode changes, no symlinks. Sizes are sane.

**Provenance is exactly what I wanted** — file, use, source, and licence per row, with Simple Icons (CC0-1.0) and lobe-icons (MIT) both compatible with this repo. That table is more than I asked for and it's the part I'd have gotten wrong on my own.

And you took the uniqueness test as written, plus pinned four specific glyphs on top. Good.

**One maintainer commit on top (`3223041`), no behavior change.** The eight ascii arms in `bar_icon` repeated the `short_name` literals verbatim:

```rust
VendorId::Kimi => "kmi",   →   VendorId::Kimi => VendorId::Kimi.short_name(),
```

Your doc comment says the fallbacks are "unique by construction" because they come from `short_name` — which is the right reasoning, but nothing in the code actually connected them. A future rename of a `short_name` would have left `bar_icon` holding the stale string and quietly broken that guarantee. Now it's structural. Same values; `cargo test`, `clippy -D warnings` and `fmt` all clean.

(Worth knowing how I found it: I wrote a throwaway test asserting every ascii `bar_icon` equals its `short_name`. It passed — so the duplication was correct *today*, which is precisely when this kind of thing is worth wiring up rather than leaving to drift.)

One note I'm recording, not acting on: MIT asks that the licence text travel with substantial portions of the work. You link it, which is the common practice for single-icon reuse and I think fine here — flagging it only so it's a decision on the record rather than an oversight.

Merging once it's refreshed past the PRs that landed while this was open.

calebdw:
Thanks!

PR #153, opened 2026-09-06
fix(openai): tolerate null optional usage collections

Swarley29:
## Summary

- Accept both JSON `null` and omitted values for Codex's optional `additional_rate_limits` and `model_usage` collections.
- Preserve strict deserialization for populated arrays and maps.
- Add a regression test for the live response shape.

## Problem

The Codex usage endpoint can return this when an account has no named limits:

```json
{
  "additional_rate_limits": null
}
```

`UsageResponse` modeled `additional_rate_limits` as `Vec<AdditionalRateLimit>` with struct-level `#[serde(default)]`. That default handles an omitted field, but it does not handle an explicitly present `null`. Deserialization therefore failed before a snapshot could be produced:

```text
openai usage response: invalid type: null, expected a sequence
```

This surfaced as an error in the Omarchy AI Usage panel for otherwise valid Codex usage responses.

## Fix

Add a small generic `de_null_default` deserializer and apply it to the two optional collection fields introduced for named limits and per-model availability. It deserializes populated values normally, but maps `null` to each collection's empty default. This keeps malformed populated data visible as schema drift instead of silently accepting it.

## Validation

- `cargo fmt --check`
- `cargo test openai::types --lib` — 29 passed
- `cargo test --lib` — 1,311 passed
- Verified against the captured live response shape that previously failed; the OpenAI entry now reaches `ready` and renders its usage window.

### DIFF
--- a/src/openai/types.rs
+++ b/src/openai/types.rs
@@ -41,10 +41,12 @@ pub struct UsageResponse {
     /// model-specific allowance. Each carries its own windows and can be the
     /// binding constraint while `rate_limit` still reads low, which is
     /// precisely when a user needs to see it.
+    #[serde(deserialize_with = "de_null_default")]
     pub additional_rate_limits: Vec<AdditionalRateLimit>,
     /// Per-model availability. `available: false` is what "Selected model is
     /// at capacity" looks like in the data — a dispatch-time refusal, not a
     /// quota, so no percentage anywhere else reflects it.
+    #[serde(deserialize_with = "de_null_default")]
     pub model_usage: BTreeMap<String, ModelUsage>,
 }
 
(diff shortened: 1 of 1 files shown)


akitaonrails:
Thank you — you were right, and fast. This was my regression: 1.11.0 added `additional_rate_limits` and `model_usage` as plain collections, I verified the feature against my own live response (which had values in both), and shipped without considering that OpenAI sends `null` for an empty one. Three of you hit it within a day.

I'm merging #151, which arrived first and also caught the nested `rate_limit_reset_credits.credits` that this one and the other don't cover. That's the only reason — your diagnosis and fix are correct and would have worked.

Your version did shape the merged result: #151 used two helpers, one for `Vec` and one for the map, and you and @wesleyramos independently wrote the single generic `T: Default + Deserialize` form. That's simpler and it's what landed, so credit where it's due. The merged fix also picks up the "still reject a wrong *type*" property — a string or a number where a collection belongs stays an error rather than being read as empty, so a real schema change can't hide behind an empty panel.

Closing as a duplicate of #151, not as wrong. A fix in 1.11.1 shortly. Sorry for the breakage, and thanks for taking the time to diagnose it properly rather than just reporting it.

PR #154, opened 2026-09-06
fix(openai): treat explicit null additional_rate_limits/model_usage as absent

wesleyramos:
## Summary

Codex usage reporting (`--vendor openai`) was failing for some accounts with:

```
API schema drift.
openai usage response: invalid type: null, expected a sequence at line 23 column 32
class: Critical
```

- `UsageResponse::additional_rate_limits` (`Vec<AdditionalRateLimit>`) and
  `UsageResponse::model_usage` (`BTreeMap<String, ModelUsage>`), added in
  v1.10.0/v1.11.0 for named limits and per-model availability, only carry
  struct-level `#[serde(default)]`.
- That attribute supplies a default **only when the key is missing** from the
  JSON. Some Codex accounts get these keys back as an explicit
  `"additional_rate_limits": null` / `"model_usage": null` instead of omitting
  them, and serde's `Vec`/`Map` deserializers reject `null` outright — which
  is exactly the "invalid type: null, expected a sequence" error.
- `ResetCreditsBlock::credits` (`Vec<ResetCredit>`) has the same exposure and
  is fixed the same way.

## Fix

Added a small `null_as_default<T: Default>` deserializer that treats an
explicit JSON `null` the same as an absent key, and applied it via
`#[serde(deserialize_with = "null_as_default")]` to the three affected
fields in `src/openai/types.rs`.

## Tests

Added two regression tests in `src/openai/types.rs`:
- `null_additional_rate_limits_and_model_usage_are_treated_as_absent`
- `null_reset_credits_detail_list_is_treated_as_empty`

`cargo test`, `cargo clippy --all-targets -- -D warnings`, and a manual run
of `./target/release/ai-usagebar --vendor openai --pretty` against a real
Codex account all pass/succeed after the fix (previously reproduced the
reported error before the change).

https://claude.ai/code/session_01WxrAHcBFDcrMFmWXrBPnN9

### DIFF
--- a/src/openai/types.rs
+++ b/src/openai/types.rs
@@ -41,11 +41,33 @@ pub struct UsageResponse {
     /// model-specific allowance. Each carries its own windows and can be the
     /// binding constraint while `rate_limit` still reads low, which is
     /// precisely when a user needs to see it.
+    ///
+    /// Some accounts get this key back as an explicit JSON `null` rather than
+    /// omitting it — struct-level `#[serde(default)]` only covers a missing
+    /// key, not a present-but-null one, so this needs its own null-tolerant
+    /// deserializer.
+    #[serde(deserialize_with = "null_as_default")]
     pub additional_rate_limits: Vec<AdditionalRateLimit>,
     /// Per-model availability. `available: false` is what "Selected model is
     /// at capacity" looks like in the data — a dispatch-time refusal, not a
     /// quota, so no percentage anywhere else reflects it.
+    ///
+    /// Same explicit-`null` caveat as `additional_rate_limits` above.
+    #[serde(deserialize_with = "null_as_default")]
     pub model_usage: BTreeMap<String, ModelUsage>,
+}
+
+/// Treat an explicit JSON `null` the same as a missing key: both mean "none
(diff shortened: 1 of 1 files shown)


akitaonrails:
Thank you — you were right, and fast. This was my regression: 1.11.0 added `additional_rate_limits` and `model_usage` as plain collections, I verified the feature against my own live response (which had values in both), and shipped without considering that OpenAI sends `null` for an empty one. Three of you hit it within a day.

I'm merging #151, which arrived first and also caught the nested `rate_limit_reset_credits.credits` that this one and the other don't cover. That's the only reason — your diagnosis and fix are correct and would have worked.

Your version did shape the merged result: #151 used two helpers, one for `Vec` and one for the map, and you and @wesleyramos independently wrote the single generic `T: Default + Deserialize` form. That's simpler and it's what landed, so credit where it's due. The merged fix also picks up the "still reject a wrong *type*" property — a string or a number where a collection belongs stays an error rather than being read as empty, so a real schema change can't hide behind an empty panel.

Closing as a duplicate of #151, not as wrong. A fix in 1.11.1 shortly. Sorry for the breakage, and thanks for taking the time to diagnose it properly rather than just reporting it.

PR #159, opened 2026-09-07
Feat/cli config override

andreperez:
## What this changes

Adds a `--config <PATH>` override to both `ai-usagebar` and
`ai-usagebar-tui`.

The option:

- Supports both `--config <PATH>` and `--config=<PATH>`.
- Works before or after subcommands.
- Applies to configuration loads, settings saves, and configuration path hints.
- Preserves platform-specific path bytes.
- Rejects missing or duplicate configuration arguments.
- Prevents accidental use of the default configuration file.

Documentation and the `Unreleased` changelog were updated with usage examples.

## Checklist

      (not applicable: this is a feature)
- [x] Docs updated where they mention configuration behavior

## Testing

Added unit tests covering:

- Flag parsing before and after subcommands
- Both supported flag formats
- Duplicate and missing flag values
- Missing configuration files
- `--` argument handling
- Non-UTF-8 platform paths
- Process-wide configuration override behavior

Validation commands still to run:

```bash
make test
cargo clippy --all-targets -- -D warnings
cargo fmt --all -- --check
cargo machete

### CODE REVIEW

**Comment by akitaonrails** [comment]
This is good — thorough, well tested (45 test functions), docs in both `README.md` and `docs/configuration.md`, and a changelog entry under the right category. Merging it.

One thing I want to record rather than block on, because it might save you code later. I tested clap directly against this crate's exact configuration (`args_conflicts_with_subcommands = true`) with a plain `#[arg(long, global = true)] config`:

```
OK    --config /tmp/a.toml
OK    --config=/tmp/a.toml
FAIL  --config /tmp/a.toml usage     ← the subcommand rejects it
OK    usage --config /tmp/a.toml
OK    usage --config=/tmp/a.toml
```

So clap handles four of the five positions natively. The `split_config_arg` extraction exists for exactly one: `--config` *before* a subcommand. That's a real capability — I'm not dismissing it — but it's ~185 lines of hand-rolled `OsString` argv walking, with its own `--` handling, `=` form, dangling-value and duplicate-flag errors, to support one ordering that clap would otherwise reject with a clear message.

If you'd rather trade that ordering for deleting the parser, `global = true` gets you the rest for free. If you think before-the-subcommand matters — and there's a decent argument that `ai-usagebar --config X account status` is the natural way to type it — then what's here is the way to get it, and it's correctly done.

Either way it works today and it's tested, so I'm not holding the PR on a design preference. Thanks for the careful work on the error cases; rejecting a dangling `--config` and a repeated one is exactly right.

**Review by andreperez — COMMENTED** [review]
Thanks for testing this against the exact config — really useful to see the five-position table laid out like that.

Context on why this exists: it started as a personal itch. Testing a new feature used to mean renaming the real config file (`~/.config/ai-usagebar/config.toml` on Unix, `%APPDATA%\ai-usagebar\config.toml` on Windows), dropping in a throwaway one, running the test, and renaming the original back — hoping I didn't forget the last step, or that the TUI Settings didn't save anything into the temporary file. `--config ./config.test.toml` kills that whole dance:

```bash
# the old way (Unix / Windows): swap the real file, swap it back
mv ~/.config/ai-usagebar/config.toml{,.bak}                             # Unix
Rename-Item "$env:APPDATA\ai-usagebar\config.toml" "config.toml.bak"    # Windows

# the new way: the real file is never touched
ai-usagebar --config ./config.test.toml usage
ai-usagebar-tui --config ./config.test.toml
```

That felt worth making public instead of keeping as a local hack — and once it's a public flag, `--config X account status` reads like the natural way to type it.

On the clap side, one scope note: `ai-usagebar-tui` doesn't use clap at all, so its `--config` parser had to be hand-written no matter what. The `global = true` comparison really only applies to `split_config_arg` in the main CLI — and there you're right that it covers 4 of the 5 positions. The one it can't fix is `--config X usage`: with `args_conflicts_with_subcommands = true`, the root command sees "I already consumed my own arg" + "a subcommand follows" as a conflict, regardless of how the arg is declared. `global = true` makes the arg visible inside each subcommand's parser, but it doesn't lift that root-level rule. The pre-parser just strips `--config` out of argv before clap ever sees it, sidestepping exactly that one case.

If that ordering ever stops paying for its ~185 lines, I'm happy to trade it for `global = true`. Either way, thanks for merging it and for writing the trade-off up so clearly.

PR #160, opened 2026-09-08
Open the Omarchy panel on an all-provider Overview

bradflaugher:
Left-click on the bar widget now opens an Overview of every configured provider (name, plan, headline usage), the same set `ai-usagebar usage` prints, instead of landing on Claude.

Click a provider row or chip for that vendor's full windows. The **all** chip (and wrapping with h/l) returns to the Overview. README screenshot recaptured from the live panel.


bradflaugher:
Opened against the wrong repository; this change belongs on bradflaugher/ai-usagebar.

PR #166, opened 2026-09-08
feat(windows): system-tray popover with a WebView2 dashboard

djalmajr:
## What this changes

On Windows the only surfaces are the TUI and `ai-usagebar --json` for hand-rolled scripts; the Waybar widget is Wayland-only and none of the desktop integrations (Omarchy, GNOME, KDE, macOS menu bar) apply. This adds the Windows counterpart of the KDE plasmoid: a NotifyIcon whose left-click opens a WebView2 popover fed in-process by the same `usage --json` report. Risk class: **a new tray host binary and its popover UI**; no vendor, cache or report behaviour changes on any OS. The auto-updater is deliberately left out and comes as a separate follow-up PR (#167).

- `ai-usagebar-tray` (new bin, `windows_subsystem = "windows"`): `src/tray/host.rs` runs the tao event loop, the `tray-icon` NotifyIcon and context menu (Refresh, Detect Providers, Open TUI, Start with Windows, Quit), a hidden always-on-top `wry` window whose height follows the page's `resize` IPC message, and a worker thread that calls `report::collect_json` / `collect_entry_json` on a 1/5/10-minute cadence. A single-instance mutex, Windows 11 rounded corners via DWM, and a blur policy that hides the popover only when focus really left the process or a mouse button is down (a taskbar that activates itself is ignored). `AIUB_TRAY_TRACE=1` logs blurs/reports/hides to `%TEMP%\ai-usagebar-tray-trace.log`.
- `src/tray/hotkey.rs`: global shortcut (`global-hotkey`) with a normalizer for the canonical `Ctrl+Shift+U` spelling; `startup.rs`: the `HKCU\…\Run` key; `tui_launch.rs`: Open TUI via `wt.exe` or `conhost.exe`; `icon.rs`: embedded 16/20/24/32/40/48 px rasters picked by `SM_CXSMICON`; `payload.rs`: the pure JSON wrapper (`HostFacts` + report → popover payload) and the icon severity, compiled and tested on every OS.
- `windows/popover/`: Vite + React + shadcn UI (dashboard, provider detail, Customize, Settings), `@dnd-kit` reordering, `unplugin-icons` provider marks, and a pure view-model `src/model.js` with a Node contract test (`popover.test.mjs`, no npm install needed). `build.rs` runs `npm run build` on Windows targets only; `dist/` is gitignored and `include_str!`'d by the host.
[...]

PR #167, opened 2026-09-08
feat(windows): in-app updates from GitHub Releases

djalmajr:
## What this changes

A Windows install of the tray is a zip the user unpacked by hand: nothing tells it a newer release exists, and nothing installs one. Linux gets that from the AUR, Nix or cargo-binstall. This teaches `ai-usagebar-tray` to poll GitHub Releases, verify the published exes against their `.sha256` sidecars, swap them into the install directory and relaunch. Risk class: **downloads and replaces binaries** — this is the one PR in the tray series that writes executables to disk, so it is split out to be reviewed on its own.

- `src/update.rs` (new, every OS, 22 unit tests): the decisions the host must not get wrong — `parse_release` (a draft or prerelease is never an update; only the safe fields of GitHub's JSON are kept), `is_newer` (numeric `X.Y.Z` compare, never a string compare, so `1.9.10` does not lose to `1.10.0`), `select_downloads` (the tray exe + sidecar pair is mandatory, CLI and TUI optional; empty or > 50 MiB assets are rejected), `parse_sha256_sidecar` (both `sha256sum` layouts, case-folded), `verify_sha256` (streamed), `stage_swap` (rename the live exe to `<name>.exe.old`, staged file into place, best-effort rollback if a later entry fails, refuses names that leave the install dir), `sweep_old`, `UpdateState { last_check_ms, snoozed_version }` in `<cache>/ai-usagebar/update.json`, and `latest_release_url_for` derived from `CARGO_PKG_REPOSITORY` so a fork updates from its own releases.
- `src/tray/update_flow.rs` (new, Windows): the `reqwest` glue — check, download to `<cache>/ai-usagebar/updates/<version>/`, verify, swap. `install` refuses to run in debug builds (`cfg!(debug_assertions)`) so a `cargo build` from a checkout never overwrites itself.
- `src/tray/host.rs`: the worker-side `Updates` machine (hourly `due()` check under `notify`/`auto`, `check(manual)`, `install`, `snooze`, `set_mode`), `WorkerCmd::{CheckUpdate, InstallUpdate, SnoozeUpdate, SetUpdates}`, `UserEvent::{Facts, Restart}`, the `set-updates` / `check-update` / `install-update` / `snooze-update` IPC arms, `sweep_old` at startup, and the relaunch: the new process is started with `AIUB_TRAY_RELAUNCH=1` and `SingleInstance::acquire_waiting` retries the mutex for up to 10 s (250 ms steps) instead of quitting because the old process has not exited yet.
- `src/tray/payload.rs`: `UpdateFact { state, version, url, error }` plus the `updates`, `update`, `update_checked_at` payload keys (sanitized).
- `[tray] updates = "auto" | "notify" | "off"` (`UpdateMode`, default `notify`) in `src/config.rs` and `config.example.toml`; a misspelled mode is a config error.
- Popover: `UpdateBanner` on the dashboard (Install / ✕ snooze), a Settings "Updates" section (mode picker, "Check for Updates" line with "Up to date · checked 33m ago", a Check Now / Update / Try Again button), a blue dot next to the footer version while a release waits, `useBusyLabel` for the "Checking…" / "Updating…" button feedback, and `model.js` helpers (`normalizeUpdate` keeps only `https://github.com/` URLs, `updateStatusLabel`, `updateBannerPending`, `updateModeLabel`, `formatAgo`) with contract tests.
- Release workflow: the Windows job also publishes the three bare `<bin>-windows-x86_64.exe` assets and a `.sha256` sidecar for each (same `sha256sum` layout as the zip's), and the release notes say what they are for. **The updater finds nothing until the first release cut with this workflow.**
- Docs: `windows/README.md` "Updates" section (with the Settings screenshot), README sentence, CHANGELOG (Added).

Screenshot (from the `djalmajr/ai-usagebar` fork): https://raw.githubusercontent.com/djalmajr/ai-usagebar/feat/windows-tray-updates/screenshots/windows-tray-settings.png

**Footprint**

| Where | Lines |
|---|---|
| `src/config.rs` (pre-existing) | +57/−11 incl. tests (`updates` field, `UpdateMode`, three tests) |
| `src/tray/host.rs` (pre-existing since the tray PR) | +295/−13 (`Updates`, IPC arms, relaunch, `acquire_waiting`) |
| `src/tray/payload.rs`, `src/tray/mod.rs`, `src/lib.rs` (pre-existing) | +53/−6, +2, +1 |
| `.github/workflows/release.yml` (pre-existing) | +33/−16 |
| `src/update.rs` (new) | 1,060 (≈ 500 code + 560 tests) |
| `src/tray/update_flow.rs` (new) | 148 |
| `windows/popover/` | `UpdateBanner.tsx` 68, `useBusyLabel.ts` 32 (new); `Settings.tsx` +73/−2, `Chrome.tsx` +12/−2, `Dashboard.tsx` +10/−1, `App.tsx` +2, `types.ts` +15, `model.js` +83, `popover.test.mjs` +79/−6 |
| `README.md`, `windows/README.md`, `config.example.toml`, `CHANGELOG.md`, 1 screenshot | +5/−3, +30/−4, +4/−1, +15 |

**Compatibility**: `updates` is one more `Option` in the `#[serde(default)]` `[tray]` table — a config without it behaves as `notify`, and a config written by this tray still loads on the previous one (unknown `[tray]` keys are tolerated there like in every other section). `update.json` is derived state beside `detect.json`; delete it to re-poll. No new crates: `sha2` and `reqwest` are already dependencies. Nothing changes for Linux/macOS: `update.rs` compiles and is tested there, `update_flow.rs` is `#[cfg(windows)]`.

Security notes for review:
- The sidecar proves integrity, not authenticity — anyone who can publish a release to the repository in `Cargo.toml` can publish a matching sidecar. There is no signature; GitHub's TLS and the repository's release permissions are the trust boundary.
- Only assets named exactly `<bin>-windows-<arch>.exe` / `.sha256` from the release JSON are considered; the banner URL is kept only when it starts with `https://github.com/`; downloads are capped at 50 MiB; the swap refuses any name that resolves outside the install directory.
- The relaunch is a plain `Command::new(<new exe>)` with `AIUB_TRAY_RELAUNCH=1` added to the inherited environment; the previous exe stays as `.exe.old` until the next start sweeps it.

**Not in this PR**
```
- Code signing / signature verification of the downloaded exes.
- Updating an install that lives in a directory the user cannot write (Program Files): the swap fails and the banner reports it; no elevation.
- An aarch64-windows asset (current_arch names it; the workflow does not build it).
- Delta/patch downloads; every update fetches the three full exes.
```

**Depends on**: #166 (included in this branch as-is; review the last commit).

## Checklist

      `cargo fmt --all -- --check` pass (`cargo test` + the four Node suites run directly; see below)
- [ ] A test that fails on `main` and passes here, for a bug fix — n/a, feature
      `config.example.toml`, `docs/`) — `windows/README.md` "Updates", README Windows paragraph, `config.example.toml` `updates`
- [x] No helper left behind without callers — `useBusyLabel` has two callers (banner, Settings); `formatAgo` one (`updateStatusLabel`); `sweep_old` / `acquire_waiting` are called from `host::run`

## Testing

Tested on **Windows 11** (24H2): the gate below plus `cargo build --bin ai-usagebar-tray`. The download-and-swap path was not exercised against a real GitHub release in this pass — the fork has no release with the bare-exe assets yet, and a debug build refuses to install by design — so what is proven here is the unit-tested decision layer (`update.rs` against temp directories: swap, rollback, sweep, sidecar parsing, version compare) and that the host compiles, wires the IPC and renders the states in the popover (Node contract test). The first release cut with this workflow is the first end-to-end run.

```
$ cargo fmt --all -- --check          # exit 0
$ cargo clippy --all-targets -- -D warnings
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 42.51s
$ cargo test
test result: ok. 1412 passed; 0 failed; 0 ignored   (lib; feat/windows-tray: 1389 → +23)
test result: ok. 0 passed   (ai-usagebar, ai-usagebar-tray, ai-usagebar-tui bins)
test result: ok. 3 passed   (tests/anthropic_e2e.rs)
test result: ok. 0 passed; 11 ignored   (tests/live.rs)
test result: ok. 3 passed   (tests/public_api.rs)
$ node gnome-extension/marker-logic.test.mjs   # marker logic tests passed
$ node kde-plasmoid/plasmoid-logic.test.mjs    # plasmoid logic tests passed
$ node omarchy/model.test.mjs                  # Omarchy model tests passed
$ node windows/popover/popover.test.mjs        # ok (exit 0; assert-only script)
$ cd windows/popover && npx tsc --noEmit       # exit 0
$ npm run build                                # dist/popover.js 459 kB, popover.css 36 kB
$ cargo build --bin ai-usagebar-tray           # Finished (build.rs ran npm run build)
```
`cargo machete` is not installed on this machine, so that step was not run.

Tests added (Rust, 23 `#[test]`s over `feat/windows-tray`):
- `update` (22, every OS): `latest_release_url_follows_the_cargo_repository_field`, `parse_release_reads_version_url_and_safe_assets`, `parse_release_accepts_a_bare_tag_and_tolerates_missing_assets`, `parse_release_rejects_prerelease_draft_and_malformed_tags`, `is_newer_compares_numerically`, `is_newer_is_false_for_anything_malformed`, `asset_name_follows_the_release_workflow`, `select_downloads_pairs_all_three_binaries_tray_first`, `select_downloads_requires_the_tray_pair`, `select_downloads_skips_an_incomplete_optional_pair`, `select_downloads_rejects_empty_and_oversized_exes`, `parse_sha256_sidecar_accepts_both_formats_and_folds_case`, `parse_sha256_sidecar_rejects_garbage`, `verify_sha256_streams_the_file_and_detects_mismatch`, `stage_swap_replaces_the_exe_and_keeps_the_old_one`, `stage_swap_installs_a_binary_that_was_not_there_before`, `stage_swap_with_a_missing_staged_file_touches_nothing`, `stage_swap_rolls_back_entries_already_swapped_when_a_later_one_fails`, `stage_swap_refuses_names_that_leave_the_install_dir`, `sweep_old_removes_only_exe_old_files`, `update_state_round_trips_and_a_corrupt_file_is_the_default`, `staging_dir_is_per_version_under_the_cache_root`
- `config`: `tray_section_rejects_a_misspelled_mode`; `tray_section_parses_and_defaults_to_notify` and `set_tray_value_creates_replaces_and_removes_keys` extended for `updates`
- `tray::payload`: `wrap_carries_shortcut_and_update_facts_sanitized` (replaces the shortcut-only version)

Node (`popover.test.mjs`): `parseHostPayload` for `updates` / `update` / `update_checked_at` incl. the defensive fallbacks (non-GitHub URL dropped, unknown state → `available`, caps), `formatAgo`, `updateStatusLabel` for every state, `updateBannerPending`, `updateModeLabel`.

All tests are hermetic: `update.rs` tests run in `tempfile` directories (`stage_swap`, `sweep_old`, `verify_sha256`, `UpdateState::{load_at,save_at}` take paths); nothing reads a real cache dir or contacts GitHub.


akitaonrails:
Reviewed this properly rather than by size. #166 is closed as superseded — this branch contains it.

**It is worth doing.** Windows was the only supported platform with no persistent surface: Linux has Waybar, GNOME, KDE and Omarchy, macOS has the menu bar, Windows had a CLI and a TUI you have to keep open. A glanceable always-on view is the point of the product, so this fills a real hole. The updater is the honest consequence — every other platform updates through a package manager and Windows has none.

**Most of it is well built**, and specifically in ways I checked rather than assumed:

- `payload.rs` is a genuinely thin wrapper over `usage --json` — pure JSON in and out, no HWND, no `$HOME`, no clock, `now_ms` injected. The architecture rule holds.
- `build.rs` returns early on non-Windows targets, so the AUR source build never touches Node, and it builds against a committed lockfile. `tao`/`wry`/`tray-icon`/`global-hotkey` are all under `cfg(target_os = "windows")`.
- The updater cannot be pointed anywhere: `SOURCE_REPOSITORY` is `env!("CARGO_PKG_REPOSITORY")`, the host is forced to `api.github.com`, owner/name are charset-validated, downloads are HTTPS-only, asset names are traversal-checked, there is a size cap, and every asset is SHA-256 verified.
- 409 assertions in `popover.test.mjs`, wired into CI as its own job.

**Two things I fixed on top (`dd79381`):**

1. **The popover grew a provider table.** `SIGN_IN_HINT` hardcoded eight providers. That is the table `CLAUDE.md` forbids in a frontend, and it had already drifted from Rust for five of the eight before shipping — Kiro's login is `kiro-cli login`, not "Sign in with kiro-cli", and grok, supergrok, cursor and antigravity have no login command at all, so the popover invented one for each. Worse, a JS object literal cannot fail to compile when a provider is added, so vendor twenty-one would silently have no hint. There is now `VendorId::sign_in_hint()` — exhaustive, so a new provider must declare one — attached per entry by the host and read by `signInHint`. Two tests cover the wiring, including that `vendor@account` resolves on the vendor half and an unknown id gets no invented hint.

2. **`clippy -D warnings` fails on `src/tray/icon.rs:117`** (`chunks_exact(4)`). That file is not Windows-gated, so the ubuntu job would have rejected this PR. It went unnoticed because this PR's CI has never run — every push here is held pending approval, so nothing has had third-party verification. Worth knowing when you read a green local run.

I also un-gated `update_flow` from `#[cfg(windows)]`. It is `reqwest` and paths with no Windows API, and the module header already states the rule — view-model helpers compile everywhere so Linux CI tests them — so its release-check logic was the one part not following it. Four mockito tests now run on Linux: a newer release, a prerelease (never offered, and not an error), the running version, and a 403 that must not leak the upstream body.

One thing for the maintainer to decide rather than for you to change: SHA-256 against a sidecar from the same release is **integrity, not authenticity** — whoever can publish a release controls both files. That is the normal model for an unsigned updater and the code is honest about it, but it means a release-account compromise becomes code execution on every Windows user's machine. The default is already `notify` rather than `auto`, which is the right call and I checked before suggesting otherwise.

Merging once the four PRs ahead of it land and this rebases cleanly.

PR #170, opened 2026-09-08
feat(macos): support all remaining CLI vendors, specific quota pools, and PATH injection

lucasliet:
## What this changes

Adds support for all remaining CLI vendors and specific quota pools in the macOS menu bar:
- **Full CLI vendor coverage**: Copilot, SuperGrok, MiniMax, Kiro, Nous Research, OpenCode Go, and Command Code are now available in the macOS menu bar.
- **Canonical metadata source**: Uses `ai-usagebar vendors --json` dynamically instead of keeping a duplicated static table in Swift. The catalog is fetched asynchronously with a watchdog timeout (`REFRESH_TIMEOUT`), generation guards against stale concurrent responses, and coalesced refresh for Overview mode.
- **Quota window extensions**:
  - Adds support for monthly quota windows (OpenCode Go, Command Code).
  - Adds support for MiniMax 5h and weekly video usage pools, including pacing markers and overview selection.
  - Distinguishes finite vs. unlimited quotas (e.g. Copilot chat & completions pools), rendering `"Unlimited"` cleanly without deceptive 0% progress rings.
- **Environment PATH injection**: Injects standard CLI paths (`/opt/homebrew/bin`, `/usr/local/bin`, `~/.cargo/bin`) into subprocess environments so CLI tools (e.g., `gh`, `kimi`, `kiro-cli`, `commandcode`) are discovered reliably.
- **Documentation**: Updated `README.md` and `macos/README.md` to reflect full binary vendor discovery via `vendors --json`, video pools, and monthly windows.

## Checklist

      (covers PATH resolution for CLI tools in GUI environment, unlimited quota rendering instead of 0% unused, and catalog arrival generation/coalescing lifecycle in `macos/ai-usagebar-tests.swift`)
      `config.example.toml`, `docs/`, `macos/README.md`)

## Testing

- Tested on macOS (Darwin arm64).
- Verified with `./macos/run-tests.sh`: all tests pass, covering snapshot parsing, quota window mapping, video pools, unlimited quota rendering, catalog contract decoding, and generation/refresh lifecycle.
- Verified `./macos/build.sh`: builds `macos/ai-usagebar-menubar` cleanly with zero warnings or errors.
- Verified `make test`: 1,543 Rust unit/integration tests and desktop frontend contract tests pass.
- Verified `cargo clippy --all-targets -- -D warnings` and `cargo fmt --all -- --check`: completely clean.
- Verified `git diff --check`: passes without whitespace errors.


akitaonrails:
Audited and merging. Thank you — this is a large, careful piece of work, and the checklist in the description was filled in honestly.

What I verified rather than took on trust:

**The placeholder contract.** The menu bar reads `--format` output *positionally* (`t(17)`…`t(26)`), which looked fragile until I saw that `FORMAT` is defined in the Swift itself — so the indices are internally consistent and the real coupling to Rust is the ~50 placeholder *names*. I extracted all fifty from `FORMAT` and confirmed every one is emitted somewhere in `src/`. Nine of those names are new here, and a name Rust does not emit would render as a silently empty field rather than an error, so this was the check worth doing. All fifty resolve.

**The new tests are actually run.** `testSubprocessEnvironment`, `testVendorCatalogContract` and `testVendorCatalogLifecycle` are all called from the harness runner, not merely defined. That is a common way for a test to look like coverage and provide none.

**CI compiled it for real.** All eight jobs pass, and the macOS job both builds the app with `swiftc` and runs `./macos/run-tests.sh`. I have no Mac, so that job is the only thing standing behind the Swift, and I would not have merged on a green Linux run alone.

Two notes, neither blocking:

- The description says this uses `vendors --json` "instead of keeping a duplicated static table in Swift". It does call `vendors --json`, but hardcoded vendor slugs in the Swift went from 27 to 36. They are per-vendor *behaviour* branches — which quota-window shape a vendor has, which balance field it uses — not the display-name table this project forbids in a frontend, so the invariant is intact. It is the description that overstates the removal, not the code that misbehaves.
- `subprocessEnvironment()` prepends `~/.cargo/bin` and `~/.local/bin` ahead of the system path. That is necessary — a macOS GUI app inherits a minimal `PATH` — and it matches what the user's own shell does, so I am not objecting. Worth knowing it is there.

The `-module-cache-path` fix in `run-tests.sh` is a good catch too.

PR #174, opened 2026-09-09
fix(omarchy): wrap provider tabs onto additional rows

### LINKED ISSUE #173: Omarchy Quattro panel: provider tabs overflow the panel edge and become unreachable
## What happened

With several providers enabled (5+), the provider tab strip in the Omarchy
Quattro panel (`omarchy/Panel.qml`) is a fixed-width, horizontal `ListView`.
Once the row of buttons is wider than the panel, the overflowing providers
are simply clipped — there is no scrollbar and no keyboard/mouse way to reach
them. They exist in `usage --json`, they're `enabled = true`, but they are
functionally invisible and unselectable in the native panel.

## Expected

The provider tabs should wrap onto additional rows once they no longer fit
on one line, the way a normal flow-layout button group would — instead of
being clipped by the panel's fixed width.

## Repro

1. Enable 5+ providers in `~/.config/ai-usagebar/config.toml`.
2. Open the Omarchy Quattro panel (left-click the bar widget).
3. The provider row is clipped at the panel's right edge; providers past
   that point cannot be selected via mouse.

## Fix

I have a small patch that swaps the `ListView` for a `Flow` (wrapping)
layout, keeping the existing `Button` delegate, styling, and keyboard/mouse
selection behavior unchanged. Opening a PR for it.

HearthCore:
Fixes #173.

## What

The Omarchy Quattro panel's provider tab strip was a fixed-width
horizontal `ListView`. With five or more providers enabled, the row
overflowed the panel's right edge and the extra entries were silently
clipped — no scrollbar, no way to reach them by mouse or keyboard.

## Fix

Swap the `ListView` for a wrapping `Flow`, keeping the exact same
`Button` delegate, styling, and selection behavior. Providers now wrap
onto additional rows once they no longer fit on one line.

`selectEntry()` no longer needs `positionViewAtIndex` — there's nothing
to scroll into view in a wrapping layout — so it calls
`providerList.forceLayout()` instead, to keep row heights correct
right after a selection change.

## Testing

- `node omarchy/model.test.mjs` — pass (added assertions for the `Flow`
  layout, the `ListView` removal, and the delegate's `width: implicitWidth`)
- Manually verified on Omarchy: with 5 configured providers (Claude,
  Codex, Copilot, Z.AI, commandcode), all five now render across two
  rows and are individually clickable; previously two of them were cut
  off past the panel edge.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,14 @@ Each release is also published at
 
 ## [Unreleased]
 
+### Fixed
+
+- **Omarchy Quattro panel:** the provider tab strip is a wrapping `Flow` again
+  instead of a fixed-width horizontal `ListView`. With five or more providers
+  enabled the list overflowed the panel's edge and the extra entries were
+  simply unreachable — no scrollbar, no way to click them. They now wrap onto
+  additional rows.
+
 ## [1.14.0] — 2026-09-08
 
 ### Added
--- a/omarchy/Panel.qml
+++ b/omarchy/Panel.qml
@@ -126,7 +126,7 @@ Panel {
     var wrapped = ((index % visibleEntries.length) + visibleEntries.length) % visibleEntries.length
     selectedEntryId = visibleEntries[wrapped].id
     persistSelection(selectedEntryId)
-    if (providerList.visible) providerList.positionViewAtIndex(wrapped, ListView.Contain)
+    if (providerList.visible) providerList.forceLayout()
     if (panelFlick) panelFlick.contentY = 0
   }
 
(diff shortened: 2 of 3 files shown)


akitaonrails:
Audited and merging. Thank you — this is a real bug and a clean fix.

I checked the thing I most expected to go wrong, which is that a wrapping layout trades horizontal clipping for vertical. It does not here: `providerList` sits in the `Column` inside the `Flickable` whose `contentHeight` tracks `column.implicitHeight`, and the panel caps at `fittedContentHeight(..., 640)` before scrolling. So extra rows grow the panel and then scroll, rather than disappearing off the bottom.

I also verified the new assertions fail against `main`'s `Panel.qml` and pass against yours — all four of them. A test that passes either way guards nothing, and source-pattern tests are especially easy to write that way by accident. These are fine, and asserting on `panelSource` is this repo's own convention for QML (there are ~60 such assertions already), since none of it is reachable from Node.

Two small notes, neither blocking:

- `providerList.forceLayout()` replaces `positionViewAtIndex(wrapped, ListView.Contain)`. The old call existed to scroll the selected tab into view; a `Flow` has nothing to scroll, so this is effectively a no-op. Keeping the selected tab visible is now handled by the `panelFlick.contentY = 0` on the next line, which is fine because the strip sits near the top of the column. Harmless as written — just noting that it is not doing what its predecessor did.
- The changelog entry is under `### Fixed`, which is right, and it explains the failure rather than only the change. That is the part people usually skip.

`make test`, the Omarchy model suite and `omarchy plugin validate` are all green on the merged result.

PR #175, opened 2026-09-09
feat: OpenCode Go pacing across CLI, TUI, report and macOS

lucasliet:
## What this changes

OpenCode Go gains pacing on its two exact windows, following the Z.AI minimal pattern:

- **CLI/placeholders** (`src/opencode_go/vendor.rs`): new `{ocg_rolling|weekly}_{elapsed,pace,pace_indicator}` plus `{session_elapsed}` / `{weekly_elapsed}` cross-vendor aliases. Window lengths are constants with recorded provenance (`ROLLING_WINDOW=5h`, `WEEKLY_WINDOW=7d`): the usage endpoint returns only `status`/`percent`/`resetsAt`, validated live against `GET https://opencode.ai/zen/go/v1/usage` (rolling message: "5-hour usage limit reached"). Rendering honors `--pace-tolerance`; the public 2-arg builder stays as a `DEFAULT_TOLERANCE` shim. Absent windows yield empty elapsed/pace (Z.AI convention); `pct`/`reset`/`status` keep the existing `—`.
- **Monthly deliberately unpaced**: its cycle follows the subscription date (`getMonthlyBounds(now, timeSubscribed)` — 28/29/31-day months), so no fixed denominator is exact. Monthly keeps `pct`/`reset`/`status` with empty pace placeholders, a glyph-free tooltip row, and a panel metric via `push_metric` (no `window_secs`) instead of `push_metric_in_window`.
- **Tooltip**: paced rows via shared `WindowRow::paced` / `push_window_with_row` (ratio glyph by default, point glyph + elapsed marker behind `--tooltip-pace-pts`); monthly row is plain (bar + percent + reset, no glyph).
- **TUI + report** (`src/tui/panels.rs`): rolling/weekly project onto `UsageWindow` and use `push_window(..., show_pacing=true)`, so the panel footnote and the `usage --json` report (Quattro/KDE/GNOME data path) carry pace automatically (monthly footnote is reset-only, no `window_secs`). No JS changes needed — `normalizeSection` renders the footnote text.
- **macOS**: session/weekly markers work with zero change via the generic `session_elapsed`/`weekly_elapsed` fields; monthly stays marker-free (no new FORMAT field).
- **GNOME**: no change — session/weekly markers flow through the existing generic `session_elapsed`/`weekly_elapsed` FORMAT fields; monthly has no slot there for any vendor (same as Z.AI MCP).

## Checklist

- [x] `CHANGELOG.md` entry under `## [Unreleased]` in the right category (not inside an already-released section)
- [x] A test that fails on `main` and passes here, for a bug fix — new-feature equivalent: the added tests (`elapsed_placeholders_follow_window_progress`, `pace_placeholders_follow_usage_vs_elapsed`, `custom_tolerance_changes_the_ratio_pace`, `monthly_keeps_reset_but_omits_pacing_until_cycle_bounds_are_known`, tooltip arrow + monthly glyph-free assertions) assert `{ocg_*_pace}`/`{session_elapsed}` placeholders that do not exist on `main` (zero occurrences of `ocg_rolling_pace`/`session_elapsed`/`ROLLING_WINDOW` in `main:src/opencode_go/vendor.rs`), so they fail there and pass here
- [x] Docs updated where they mention what changed (`README.md`, `config.example.toml`, `docs/`) — `docs/format-placeholders.md` gains the OpenCode Go pace family (rolling/weekly; monthly documented as reset-only); no config surface changed

## Testing

- `make test` (cargo test + GNOME, KDE, Omarchy and Windows contract suites) — green (1551 Rust tests)
- `cargo fmt --all -- --check` — clean
- `cargo machete` — no unused dependencies
- `./macos/run-tests.sh` — all tests passed
- Live check against the real API: CLI `{ocg_rolling_pace}`/`{session_elapsed}` expand (`1% ↓ el:11`), monthly pace placeholders expand empty with reset kept (`mo:[][] 23d 1h`); report shows rolling/weekly footnotes with pace + `window_secs` (18000/604800) and monthly as reset-only with no `window_secs`
- Tested on macOS (Swift harness ran locally); Linux paths covered by `make test`. Windows/macOS CI builds will exercise the rest.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,13 +23,15 @@ Each release is also published at
   `/usr/local/bin`, and `~/.cargo/bin` into subprocess environments for vendor
   tools.
 
-- **OpenCode Go pacing.** The rolling (5h), weekly (7d), and monthly (30d)
-  windows now expose `{ocg_*_elapsed}`, `{ocg_*_pace}`, and
-  `{ocg_*_pace_indicator}` placeholders plus `{session_elapsed}` /
-  `{weekly_elapsed}` aliases, pace arrows in the Waybar tooltip, paced rows in
-  the TUI panel and `usage --json` report footnotes, and the monthly elapsed
-  marker in the macOS menu bar. Window lengths are constants: the usage
-  endpoint reports only `percent` and `resetsAt`.
+- **OpenCode Go pacing.** The rolling (5h) and weekly (7d) windows now expose
+  `{ocg_rolling|weekly_elapsed}`, `{ocg_rolling|weekly_pace}`, and
+  `{ocg_rolling|weekly_pace_indicator}` placeholders plus `{session_elapsed}`
+  / `{weekly_elapsed}` aliases, pace arrows in the Waybar tooltip, and paced
+  rows in the TUI panel and `usage --json` report footnotes. The monthly
+  window keeps its reset countdown but is not paced: its cycle follows the
+  subscription date, so no fixed length is exact and no `window_secs` is
+  published for it. Window lengths are constants: the usage endpoint reports
+  only `percent` and `resetsAt`.
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,14 @@ Each release is also published at
   `/usr/local/bin`, and `~/.cargo/bin` into subprocess environments for vendor
   tools.
 
+- **OpenCode Go pacing.** The rolling (5h), weekly (7d), and monthly (30d)
+  windows now expose `{ocg_*_elapsed}`, `{ocg_*_pace}`, and
+  `{ocg_*_pace_indicator}` placeholders plus `{session_elapsed}` /
+  `{weekly_elapsed}` aliases, pace arrows in the Waybar tooltip, paced rows in
+  the TUI panel and `usage --json` report footnotes, and the monthly elapsed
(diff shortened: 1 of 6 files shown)


akitaonrails:
Audited and merging. This is careful work and the description is accurate, which is not always the same thing.

The part I want to call out is the monthly window. You left it unpaced on the grounds that the cycle follows the subscription date, so 28/29/31-day months mean no fixed denominator is exact — and you used `push_metric` rather than `push_metric_in_window` so `window_secs` is withheld from machine-readable frontends, with a comment saying why.

That is exactly the discipline this project needed and did not always have. In #162 the Cursor billing cycle did the opposite: it fell back to an assumed 30 days when the API omitted `billingCycleStart`, so a guessed month reached frontends as an exact window and was paced as one. I had to change `cycle_window()` to return `Option` to stop it. You arrived at the right answer unprompted, and documented the reasoning in the code where the next person will find it.

Verified rather than assumed:

- `ROLLING_WINDOW = 5h` and `WEEKLY_WINDOW = 7d` are constants with recorded provenance, which is the bar for reporting `window_secs` at all.
- The "zero macOS change" claim holds: the menu bar's `FORMAT` already reads `{session_elapsed}` and `{weekly_elapsed}`, and this PR touches no file under `macos/`. Pacing arrives there through the generic aliases.
- Gate is green — 1570 tests, clippy `-D warnings` and fmt clean.

Merging as-is. No maintainer commits needed on this one.

PR #176, opened 2026-09-09
feat(ollama): add native Ollama Cloud usage vendor

rhgs:
## Summary

Adds native **Ollama Cloud** usage monitoring to ai-usagebar — a new vendor that reports 5-hour session and weekly rolling usage windows from `ollama.com/api/usage`, the same undocumented route the official ollama.com/settings page calls.

## Motivation

Ollama Cloud users have no CLI or desktop widget to track their plan usage percentage. The official settings page shows session/weekly usage as a fraction of the plan limit, per-model request counts, and a 4-week activity cost — but only in the browser. This PR brings that same data into the TUI, Waybar widget, and `usage --json` report alongside every other supported vendor.

## Approach

- **New provider module** `src/ollama/` with `fetch.rs`, `types.rs`, `vendor.rs`, and `mod.rs`
  - Bearer-auth GET to `https://ollama.com/api/usage` (10s timeout, 15s file lock)
  - `usage` is a fraction `[0,1]` — clamped to `0..=100` for display
  - Per-model `request_count` arrays for session and weekly windows
  - `activity.cost` is a dollar string; `activity.period` is a labeled span
  - The cache stores only the projected snapshot — the raw body (and the Bearer key with it) is never written to disk
- **Vendor wiring** across all 14 integration points: `VendorId` enum, `config.rs` (`OllamaConfig`), `catalog.rs`, `detect.rs`, `active.rs`, `widget/cli.rs`, `widget/run.rs`, `tui/app.rs`, `tui/settings.rs`, `tui/panels.rs`, `usage.rs`
- **Credential model**: a Bearer API key minted at `ollama.com/settings/keys`, passed via `OLLAMA_API_KEY` env var or inline `[ollama] api_key`. The CLI's Ed25519 registry key in `~/.ollama` is **never read** — it is for model registry, not quota. The local daemon at `127.0.0.1:11434` has no quota route and is never contacted.
- **Plan label**: the API does not report one; `plan` in config is purely a display name for tooltips (defaults to `"pro"`).
- **Settings overlay fix**: `KEY_VENDORS` in `tui::settings` were excluded from the primary list when neither an inline `api_key` nor an env var resolved at startup. This patch treats a non-empty env var as sufficient (matching the behavior of env-only key vendors like DeepSeek/Kimi), and `save_to_path` flips `enabled = true` for any `KEY_VENDORS` primary — not just Copilot.

## Files changed

**New (untracked → added):**
- `src/ollama/{mod.rs, types.rs, fetch.rs, vendor.rs}` — provider module
- `tests/fixtures/ollama/good_full.json` — real captured response shape
- `docs/ollama-setup.md` — setup guide (Bearer key, troubleshooting)
- `docs/windows-build.md` — Windows build prerequisites (rustup, MSVC, NASM)
- `DEVELOPMENT.md` — development guide (prerequisites, build, test, lint, config)

**Modified (20 tracked files, +394/−12):**
- `src/lib.rs`, `src/usage.rs`, `src/vendor.rs`, `src/config.rs`, `src/catalog.rs`, `src/detect.rs`, `src/active.rs` — vendor registration + types
- `src/widget/cli.rs`, `src/widget/run.rs` — CLI dispatch + `ollama_output`
- `src/tui/app.rs`, `src/tui/panels.rs`, `src/tui/settings.rs` — TUI build_outcome, panel sections, settings overlay
- `tests/live.rs` — `ollama_live` smoke test (ignored by default)
- `config.example.toml`, `docs/configuration.md` — `[ollama]` config block
- `docs/format-placeholders.md` — `{oll_*}` placeholder reference
- `docs/vendor-endpoints.md` — endpoint matrix + stability note
- `README.md` — reference guides + Ollama pointer
- `CHANGELOG.md` — `### Added` entry under `[Unreleased]`
- `.gitignore` — `config.*.toml` scratch-file rule

## Pre-PR gate (all green)

- [x] `cargo test` — 0 failures (ollama_live present, ignored as live API)
- [x] `cargo clippy --all-targets -- -D warnings` — clean
- [x] `cargo fmt --all -- --check` — clean
- [x] `cargo machete` — no unused dependencies

## Live smoke test

Verified against the real API with a Bearer key minted at ollama.com/settings/keys:
- Session window: 100% (plan limit reached)
- Weekly window: 26%
- 5 session models, 6 weekly models (kimi-k3, deepseek-v4-flash:0731, minimax-m3, glm-5.3-flash, gpt-oss:120b, qwen3.5:397b)
- Activity cost: $0.00000 (last 4 weeks)

## Security notes

- The Bearer key is sent only to `ollama.com` over HTTPS and is never logged, cached, or written to config
- The raw API response body is never persisted to the cache — only the projected `OllamaSnapshot` (which contains no credentials)
- The CLI's local Ed25519 key (`~/.ollama/id_ed25519`) is registry-only and explicitly never read

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,6 +12,11 @@
 .env.local
 *.credentials.json
 
+# Per-vendor config scratch files (config.<vendor>.toml) — never commit.
+# Only config.example.toml and the user's own config.toml are tracked.
+config.*.toml
+!config.example.toml
+
 # Snapshot review files from `cargo insta`
 *.snap.new
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,35 @@ Each release is also published at
   simply unreachable — no scrollbar, no way to click them. They now wrap onto
   additional rows.
 
+- **Settings overlay can pick env-only key vendors.** `KEY_VENDORS` in
+  `tui::settings` were excluded from the primary list whenever neither an
+  inline `api_key` nor a non-empty env var resolved at startup, so a fresh
+  install that only ever exports `OLLAMA_API_KEY` (or any other key
+  vendor's env var) could not select the matching tab in the TUI to
+  flip `enabled = true` without first editing the TOML by hand. The
+  overlay now treats a present env var as a sufficient signal that the
+  vendor is reachable, surfaces it in the primary list and writes
+  `enabled = true` on save like the inline-key path did.
+
+### Added
+
+- **Ollama Cloud vendor.** `ollama.com/api/usage`, the quota route the
+  official settings page itself uses, behind a Bearer key minted at
+  <https://ollama.com/settings/keys>. The local daemon at
+  `127.0.0.1:11434` has no quota route, and the Ed25519 key the `ollama`
+  CLI keeps in `~/.ollama/id_ed25519` is the registry's signing key, not
+  a quota credential — the widget never reads it. The native Rust
+  provider adds an `Ollama Cloud` tab, a `[ollama]` config block
+  (disabled by default, opt in with `enabled = true` or via the TUI
+  Settings overlay), `{oll_session_pct}` / `{oll_weekly_pct}` placeholders
+  with the usual pace and reset aliases, a per-model breakdown of the
+  five heaviest models in each window in the tooltip, and a
+  `usage --json` entry keyed `ollama`. Plan label comes from config; the
+  API itself does not report one. `tests/fixtures/ollama/good_full.json`
+  pins the real shape, and `tests::live::ollama_live` is the live smoke
+  against the real API. The cache stores the projected snapshot only —
+  the raw body, and the Bearer key with it, is never written to disk.
+
 ## [1.14.0] — 2026-09-08
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,23 @@ Each release is also published at
   `/usr/local/bin`, and `~/.cargo/bin` into subprocess environments for vendor
   tools.
 
+- **Ollama Cloud vendor.** `ollama.com/api/usage`, the quota route the
+  official settings page itself uses, behind a Bearer key minted at
+  <https://ollama.com/settings/keys>. The local daemon at
+  `127.0.0.1:11434` has no quota route, and the Ed25519 key the `ollama`
+  CLI keeps in `~/.ollama/id_ed25519` is the registry's signing key, not
+  a quota credential — the widget never reads it. The native Rust
+  provider adds an `Ollama Cloud` tab, a `[ollama]` config block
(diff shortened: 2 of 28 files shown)


lucasliet:
Recently, I paired the providers with the macOS module in [PR !170.](https://github.com/akitaonrails/ai-usagebar/pull/170/changes)

It would be great for this provider addition to also be reflected in the Swift classes, following the same pattern.

akitaonrails:
Audited. This is a well-built vendor — it follows the module layout, goes through `outcome::fallback` rather than touching `Cache::fallback_payload`, registers in `catalog`/`detect`/`vendor` exhaustively, documents the endpoint in `docs/vendor-endpoints.md`, keeps the live test behind `#[ignore]`, and the `.gitignore` addition for `config.<vendor>.toml` scratch files is a good instinct. Thank you.

I pushed one commit on top (`295b778`), and it is worth explaining because the bug is invisible on most machines including CI's.

**`SettingsState::from_config` began reading the ambient environment.** The new block offering a key-only vendor whose env var is already exported is good behaviour — a user who has `OLLAMA_API_KEY` in their shell should be able to select Ollama in the overlay without hand-editing `config.toml`. But it made `from_config_offers_enabled_vendors_only` depend on the machine running the test. It failed immediately here, because I have those variables set.

I proved the dependency rather than assuming it:

```
env -u OLLAMA_API_KEY -u KIMI_API_KEY … cargo test from_config_offers_enabled_vendors_only  → ok
OLLAMA_API_KEY=x                        cargo test from_config_offers_enabled_vendors_only  → FAILED
```

That matters more than a flaky test, because of how this project is packaged. `CLAUDE.md` puts it plainly: *a `#[test]` must never … branch on an ambient env var — the AUR `check()` runs `cargo test` during `makepkg`*. So `makepkg` would have failed for every Arch user with `OLLAMA_API_KEY` exported — which is precisely the set of people who would want this vendor. It would have looked like a broken package, not a test bug.

The fix is `from_config_with(cfg, env_set)`, matching the seam pattern used elsewhere (`Cache::at`, `creds::read_from`, `Cli::resolve_vendor_with`). The existing test passes `|_| false`; a new one, `a_key_vendor_with_its_env_var_exported_is_offered`, pins your behaviour with an injected lookup so it is tested rather than merely working. Verified by running the whole suite with four vendor keys exported: 1581 passed either way.

Also — thank you for fixing the changelog category duplication yourself in `5febb95` while I was looking at this. The earlier CI failure was the `no_changelog_section_repeats_a_category` guard firing on the *merge* result rather than your head, which is exactly the case it exists for.

Merging once this run is green.

rhgs:
Thanks for the review and feedback! 
 I'm glad to contribute to the project. If you see anything that could be improved or adjusted, I'd be more than happy to make the necessary changes.
 Thanks again for taking the time to review the PR and share your suggestions.

PR #184, opened 2026-09-10
fix(tray): use correct wt syntax for Open TUI launch

rhgs:
## Summary

The tray icon's "Open TUI" menu item was launching `wt.exe` with the `-e` flag:
```rust
Command::new("wt.exe").args(["-e", &tui]).spawn()
```

**`-e` is not a Windows Terminal flag** — it belongs to wezterm. Windows Terminal
rejects it, prints its Usage/help page, and exits immediately. The TUI never
starts; the user sees a brief flash of the `wt` help text and nothing else.

## Fix

Use the correct Windows Terminal syntax: `wt new-tab -- <command>`.

```rust
Command::new("wt.exe")
    .args(["new-tab", "--", &tui])
    .spawn()
```

The `conhost.exe` fallback is kept unchanged as the last resort.

## Verification

- `cargo check --all-targets` — clean
- `cargo fmt --check` — clean
- `cargo build --bin ai-usagebar-tray` — clean
- Manual test: clicking "Open TUI" in the tray now opens a Windows Terminal tab
  running the TUI correctly.

### DIFF
--- a/src/tray/tui_launch.rs
+++ b/src/tray/tui_launch.rs
@@ -7,7 +7,14 @@ pub fn open() {
     let Some(tui) = resolve_tui() else {
         return;
     };
-    if Command::new("wt.exe").args(["-e", &tui]).spawn().is_ok() {
+    // `wt -e` is not a real Windows Terminal flag (it belongs to wezterm);
+    // wt rejects it, prints its Usage page, and exits — the TUI never starts.
+    // The correct syntax is `wt new-tab -- <command>`.
+    if Command::new("wt.exe")
+        .args(["new-tab", "--", &tui])
+        .spawn()
+        .is_ok()
+    {
         return;
     }
     let _ = Command::new("conhost.exe").arg(&tui).spawn();


akitaonrails:
Correct diagnosis and correct fix — merging.

Verified the claim rather than taking it: `-e` is wezterm's flag, and Windows Terminal's own CLI is `wt new-tab -- <command>`. The reason this looked like it worked is in your summary and worth restating for the archive: `spawn()` reports only that the process *started*. `wt.exe` started fine, rejected the flag, printed its usage page and exited — so the launcher saw `Ok`, treated it as success, and never fell through to the next terminal in the chain. A silent failure that reports success is the worst shape for this kind of bug.

One thing I added on top (`2716847`): a `CHANGELOG.md` entry under `### Fixed`. This PR had none, and the fix is user-visible — "Open TUI" did nothing on every Windows install. The release process reads those category headings to pick the next version number, so an unlisted fix is invisible to it as well as to users.

No test, which I am not going to hold this on: `src/tray/tui_launch.rs` is `#[cfg(windows)]`, so nothing on Linux or macOS compiles it, and the failure mode is an external program's argument parsing rather than logic we own. That is a real coverage gap in the tray's launcher paths generally, not something this three-line fix should be asked to solve.

PR #185, opened 2026-09-10
fix(macos): default Ollama Cloud to disabled, matching Rust

lucasliet:
## What this changes

Follow-up to #176: Ollama Cloud shipped in v1.15.0 with full Rust wiring, but the macOS menu bar's `defaultEnabled("ollama")` fell through to `true` while Rust ships `[ollama] enabled = false` (opt-in, `src/config.rs`). Now returns `false`, matching Rust.

- `macos/ai-usagebar-menubar.swift`: `ollama` added to the opt-in list; two comments reworded to not overstate their rules ("balance vendors" → "vendors"; picker comment now describes the catalog filter instead of asserting the default rule).
- `macos/ai-usagebar-tests.swift`: `ollama` pinned in `testDefaultEnabled` + new parse test proving Session/Weekly (with pace `elapsed`) render through the generic `parse()` path — no vendor-specific case or FORMAT change needed.
- `CHANGELOG.md`: entry under `[Unreleased]` → `Fixed`.

## Checklist

- [x] `CHANGELOG.md` entry under `## [Unreleased]` in the right category (not inside an already-released section)
- [x] Docs updated where they mention what changed (`README.md`, `config.example.toml`, `docs/`)

Docs note: no doc enumerates the macOS `defaultEnabled` list, so nothing else mentions what changed; `docs/configuration.md` already documents `[ollama]` as opt-in.

## Testing

- `macos/run-tests.sh` on macOS: all tests passed, including the 9 new Ollama asserts.
- Old source (`main:macos/ai-usagebar-menubar.swift`) + new tests: `ollama defaults disabled (opt-in)` fails with `got true, expected false` — 1 test FAILED, as expected.
- `make test`: 1564 passed, 0 failed (+ GNOME/KDE/Windows/Omarchy contract suites green).
- `cargo clippy --all-targets -- -D warnings`: clean. `cargo fmt --all -- --check`: clean. `cargo machete`: no unused deps.
- Tested on macOS (Swift harness runs natively here).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,16 @@ Each release is also published at
 
 ## [Unreleased]
 
+### Fixed
+
+- **macOS menu bar knew the wrong default for Ollama Cloud.**
+  `defaultEnabled("ollama")` fell through to `true` while Rust ships
+  `[ollama] enabled = false` (opt-in), so the menu bar could treat Ollama
+  as enabled when the config omits the flag. It now returns `false`,
+  matching `src/config.rs`, and the Swift contract test pins it alongside
+  the other opt-in vendors. Ollama's Session/Weekly bars already rendered
+  through the generic `parse()` path — no format change.
+
 ## [1.15.0] — 2026-09-10
 
 ### Added
--- a/macos/ai-usagebar-menubar.swift
+++ b/macos/ai-usagebar-menubar.swift
@@ -1365,13 +1365,13 @@ func addAccountScript(binary: String, label: String, desktop: Bool) -> String {
 }
 
 /// Rust defaults (`src/config.rs`): the OAuth/api-key vendors that ship enabled,
-/// versus the opt-in balance vendors that default to disabled. An omitted
+/// versus the opt-in vendors that default to disabled. An omitted
 /// `[vendor].enabled` must reproduce these, not silently enable everything.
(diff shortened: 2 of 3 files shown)


akitaonrails:
Verified against Rust rather than the description, and it checks out: `OllamaConfig::default()` ships `enabled: false`, `vendors --json` reports `enabled: false`, and the menu bar's `defaultEnabled` was the only place saying otherwise. Merging.

Worth naming the shape of this bug, because it is the third time this pattern has cost us something recently. `defaultEnabled` in Swift is a hand-maintained copy of a rule Rust already owns and already exports. Adding a vendor to Rust does not fail any compile or test in Swift, so the copy silently drifts — here a provider that Rust deliberately makes opt-in was on by default in the menu bar, which for a key-only vendor means a card that can only fail until the user adds a key.

#170 moved the *catalog* to `vendors --json` for exactly this reason. The defaults are the remaining half: `vendors --json` already carries `enabled` per provider, so `defaultEnabled` could read it instead of listing slugs. Not asking you to do that here — this PR is the right size for the bug — but it is the follow-up that would stop the next one.

The new parse test earning its place is the part I liked: proving Session/Weekly with pace render through the generic `parse()` path is what shows no vendor-specific case or FORMAT entry was needed, which is the claim that would otherwise be taken on trust.

PR #189, opened 2026-09-11
fix(cursor): show on-demand spend and limit

### LINKED ISSUE #188: Show Cursor on-demand spend and limit
## Problem

Cursor Enterprise accounts expose numeric on-demand spending in `GET /api/usage-summary`, but the Cursor integration only deserializes `onDemand.enabled`. The Omarchy panel therefore says that on-demand usage is enabled without showing the amount used or its configured limit.

A redacted response shape is:

```json
{
  "teamUsage": {
    "onDemand": {
      "enabled": true,
      "used": 1955,
      "limit": 35000,
      "remaining": 33045
    }
  }
}
```

Values are expressed in cents.

## Expected behavior

When the fields are available, include an On-Demand row in Cursor usage sections, for example:

```text
On-Demand    $19.55 / $350
```

Keep compatibility with responses and cached snapshots where the numeric fields are absent.

MartinGuerraGit:
Closes #188

## What this changes

- Deserialize Cursor's optional `onDemand.used` and `onDemand.limit` cent values for personal and team/Enterprise response shapes.
- Preserve the values in account-scoped caches and show an `On-Demand` spend/limit row in usage sections.
- Keep older responses and caches compatible when numeric on-demand fields are absent.

## Checklist

- [x] Docs updated where they mention what changed (`docs/vendor-endpoints.md`)

## Testing

- `make test` — 1,544 Rust tests plus desktop and Omarchy contract suites passed.
- `cargo clippy --all-targets -- -D warnings` — passed.
- `cargo fmt --all -- --check` — passed.
- Live-tested on Linux with a Cursor Enterprise account; the Omarchy panel reported the current spend and configured limit.

image


akitaonrails:
Audited and merging. Closes #188.

I checked this against a specific past failure rather than in the abstract. v1.11.0 shipped a regression where `#[serde(default)]` covered a *missing* field but not a present-but-`null` one — OpenAI sends `null` rather than `[]` — and three people reported it inside a day. Your new fields use the same attribute, so it was worth establishing whether the same trap applies.

It does not: `Option<i64>` decodes `null` itself, unlike the `Vec`/`BTreeMap` that broke before. I proved that rather than reasoning about it and kept the check as a test (`on_demand_tolerates_missing_null_and_present`), because this is the one module where getting it wrong has already cost a release, and the next person adding an optional field there should inherit the answer instead of the argument.

Two other things that were right and are easy to get wrong:

- `.filter(|v| *v >= 0)` on both cent values. A negative from an unexpected response becomes absent rather than a nonsense row.
- Rendering through `usage::fmt_minor(..., Some("USD"))` rather than `format!("${:.2}")`. That matters here — this project's money formatting had regrown into four disagreeing copies once, so there is one source and you used it.

Merging with the added test on top.

PR #190, opened 2026-09-11
feat(omarchy): top bar usage-window picker

mickeiik:
## What this changes

Adds a **Top bar usage window** picker to the Omarchy Quattro widget. The
settings page gains a dropdown (`auto` / `session` / `weekly` / `monthly`), and
the same key is settable from the CLI:

```
omarchy bar set akitaonrails.ai-usagebar barWindow session
```

`barWindow` pins the bar label — and the tooltip/hero echo — to one quota
window instead of always showing the highest percent:

- `session` — the 5-hour / rolling window
- `weekly` — the 7-day window
- `monthly` — the monthly pool where one exists
- `auto` (default) — the historical highest-percent behavior; existing
  `shell.json` entries are unchanged

A vendor that does not offer the pinned window falls back to the highest
percent rather than blanking the bar. Panel rows and alert state always follow
the worst window, matching KDE's `isAlarming` and Waybar's class; the pin is
display-only.

Classification consumes the report's existing `window_secs` first; the two
exact sizes are named after the Rust constants that publish them
(`openai/types.rs` `SESSION_WINDOW_SECS`/`WEEKLY_WINDOW_SECS`,
`opencode_go/vendor.rs` and `kimi/vendor.rs` `ROLLING_WINDOW`/`WEEKLY_WINDOW`,
`anthropic/types.rs` `SESSION`/`WEEKLY`). Bounded label matching only covers
older reports that predate `window_secs`. A vendor stating any other length
falls through to labels rather than being forced into one of these classes. If
you would rather this lived in the report as a canonical window tag, I'm happy
to move it.

## Checklist

      `cargo fmt --all -- --check` pass (also `cargo machete`,
      `make changelog-check`, `omarchy plugin validate .`)
- [x] Docs updated (`README.md`, `omarchy/README.md`)

## Testing

Gate on `8a58593`, Linux (Omarchy Quattro):

```
make test                                   # 1583 cargo tests passed, 0 failed, 13 ignored
                                            #   + GNOME, KDE, Windows popover, Omarchy suites
cargo clippy --all-targets -- -D warnings   # clean
cargo fmt --all -- --check                  # clean
cargo machete                               # no unused dependencies
make changelog-check                        # released sections intact
omarchy plugin validate .                   # exit 0
```

- Exercised live in a running Quattro session: selecting each window in the
  dropdown updates the bar label, tooltip and hero immediately; the choice
  persists.
- **Fails on `main`:** copy this branch's `omarchy/model.test.mjs` into a
  `main` worktree and run `node omarchy/model.test.mjs` — it fails at line 32
  (`barWindow` default is `undefined`, expected `'auto'`). The behavioral
  assertions below it call `Model.normalizeBarWindow`/`metricMatchesWindow`,
  which don't exist on `main`.
- **Real-report check:** piped `ai-usagebar usage --json` through
  `omarchy/Model.js` in Node — `openai` (Codex 5h 53 / weekly 60) resolves
  auto 60, session 53, weekly 60; `opencode-go` (Rolling 6 / Weekly 20 /
  Monthly 82) resolves auto 82, session 6, weekly 20, monthly 82; a vendor
  with no monthly pool (openai) falls back to 60% as documented.
- **Fixtures use the exact labels/sizes `src/tui/panels.rs` emits:** Z.AI
  `MCP tools (monthly)` (30d), MiniMax `Text`/`Video` (5h/24h/7d), SuperGrok
  `Monthly Build credits` with no `window_secs`, plus a manifest-options ↔
  normalizer round-trip guard.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,20 @@ Each release is also published at
 
 ## [Unreleased]
 
+### Added
+
+- **Omarchy top bar usage-window picker.** The Quattro settings page gains a
+  **Top bar usage window** dropdown (`auto` / `session` / `weekly` /
+  `monthly`), also settable with
+  `omarchy bar set akitaonrails.ai-usagebar barWindow session` that pins the
+  bar label to one quota window instead of always showing the highest percent;
+  the tooltip and panel hero echo the pinned value. `session` pins the 5-hour
+  window, `weekly` the 7-day window, and `monthly` the monthly pool where one
+  exists; `auto` keeps the historical highest-percent behavior and is the
+  default, so existing installs are unchanged. A pinned window a vendor does
+  not offer falls back to the highest percent rather than blanking the bar.
+  Panel rows and alert state still follow the highest percent regardless.
+
 ## [1.15.0] — 2026-09-10
 
 ### Added
--- a/README.md
+++ b/README.md
@@ -155,7 +155,10 @@ wheel to switch providers. In QML settings, turn off **Show usage value in the
 top bar** for an icon-only widget; the panel and tooltip keep the full details.
 Turn on **Show provider name in the top bar** to prefix the entry with the same
 three-letter code Waybar's `{vendor_short}` prints, so a bar cycling several
-providers says which one it is showing.
+providers says which one it is showing. Use **Top bar usage window** to pin the
+bar to one quota window — auto (highest), 5-hour, weekly, or monthly — instead
+of always showing the highest percent; the tooltip and panel hero echo the
+pinned value while panel rows and alert state still follow the highest quota.
 
 The source-built `ai-usagebar` AUR package can replace `ai-usagebar-bin` in
 the first command.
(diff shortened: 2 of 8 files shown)


akitaonrails:
Audited and merging.

The design choice I want to credit: the picker falls back to the highest window when a vendor lacks the pinned one, rather than showing an empty or zeroed row. A bar item that silently reads 0% because the selected window does not exist for that provider would be worse than not offering the setting — and with providers now ranging from single-window to four-window, "the window you pinned does not apply here" is the common case, not the edge.

Checked the things that bite in this repo:

- `manifest.json` gains the `barWindow` default and its schema entry, and leaves `version` alone. That field belongs to the release process, and a PR bumping it is how two earlier releases got into trouble.
- `omarchy plugin validate .` passes on the merged result, and the model suite is green — the enum's four options match what `Model.js` actually branches on, so a value from the CLI (`omarchy bar set ... barWindow session`) cannot land somewhere unhandled.
- The description on the schema entry explains the fallback, which is what a user reading the settings page needs rather than a restatement of the label.

Thanks for wiring it through both the QML settings page and the CLI key — the same setting reachable two ways with one source of truth is the right shape.

PR #191, opened 2026-09-11
refactor(macos): drop the hand-maintained enabled-defaults copy

lucasliet:
## What this changes

Follow-up to the suggestion in #185's review: `vendors --json` already carries `enabled` per provider, so the menu bar should not keep its own list of which slugs default on.

What the sweep found: #170 had already moved **every live enabled decision** to the catalog field (`vendorEntries`, `SettingsView.vendors`) — orphaning `defaultEnabled` and its helpers. The copy #185 fixed was consulted by nothing; its Ollama disagreement was latent (the catalog migration landed before the Ollama vendor existed), never user-visible. So the suggestion lands as a deletion, not a repoint:

- Deletes the orphaned family: `defaultEnabled`, `vendorEnabled`, `configEnabledTOML`, `configHasApiKeyTOML`, `vendorConfigured` — zero callers in either Swift file (grep-verified; the harness compiles both files as one module, so a missed caller would fail the build).
- Repoints the contract: `testDefaultEnabled` (which pinned the hand copy against itself) is gone; `testVendorCatalogContract` now asserts the `enabled` field decides (copilot enabled / supergrok opt-in).
- Pins the other side of the boundary: the Rust wire test in `src/catalog.rs` now asserts `enabled`/`configured`/`short_name`. Before this, a serde-level rename of `enabled` would pass every gate and silently empty the menu bar's selector (`parseVendorCatalog`'s `try?` → nil, stderr discarded).
- Comments that re-enumerated the defaults (the `SettingsView.vendors` prose, the `tomlValueInText` bare-boolean example) now name no vendor and use a key a live caller actually reads (`show_default_account`).
- Rewords the two `[Unreleased]` entries (this change and #185's) so the shipped story is accurate about the drift being latent.

Net: a provider added in Rust reaches the menu bar with its own default, no Swift change — the #185 bug class becomes unrepresentable in Swift.

**Not here, but found while sweeping:** the GNOME extension never migrated to the catalog — `prefs.js` still keeps `VENDOR_AUTH` (6 of 21 providers) plus a hand-written TOML reader, and the module comment in `src/catalog.rs` reads as if both frontends had migrated. That is the natural next PR of this same shape.

## Checklist

- [ ] A test that fails on `main` and passes here, for a bug fix — N/A:
      behavior-preserving removal of dead code, so no test distinguishes
      `main` from this branch by design; the safety argument is the caller
      inventory plus the macOS CI job
      `config.example.toml`, `docs/`) — nothing mentions the deleted
      symbols (grep-checked); no behavior changed for docs to describe

## Testing

- `./macos/run-tests.sh` (exit 0) and the CI's exact app-build command
  (`swiftc -O -parse-as-library -o /tmp/ai-usagebar-menubar`) on a real
  macOS host — Apple Swift 6.3.3, arm64, via ssh
- Linux: `make test` (1565 Rust tests + the GNOME/KDE/Omarchy/Windows Node
  suites), `cargo clippy --all-targets -- -D warnings`, `fmt --check`,
  `machete`, `make changelog-check`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,18 @@ Each release is also published at
 
 ### Changed
 
+- **macOS menu bar keeps no second copy of Rust's enabled defaults.**
+  The `defaultEnabled` slug list is deleted, together with the helpers the
+  catalog migration had orphaned (`vendorEnabled`, `configEnabledTOML`,
+  `configHasApiKeyTOML`, `vendorConfigured`). Every enabled decision in the
+  menu bar had already read the catalog's `enabled` field from
+  `vendors --json` since the #170 migration — which is why the slug list's
+  disagreement about Ollama Cloud (#185) was latent, never user-visible —
+  and now nothing else exists to drift: the Rust wire test pins the
+  `enabled` field name, and the Swift contract test pins that the field
+  decides. A provider added in Rust reaches the menu bar with its own
+  default, no Swift change needed.
+
 - **Omarchy install is one paste, and the marketplace card says what it needs.**
   The plugin is the display frontend; it reads the `ai-usagebar` binary, which
   installs through a different manager (the binary is a system package, the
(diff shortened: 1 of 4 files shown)


akitaonrails:
Merging — and thank you for doing the sweep rather than the repoint I suggested, because it corrects something I got wrong.

On #185 I wrote that the Ollama default divergence meant "a card that can only fail until the user adds a key." **That was wrong, and your investigation is what shows it.** I verified the orphan claim myself rather than taking it: `defaultEnabled`'s only non-test caller is `vendorEnabled`, `vendorEnabled` has zero callers, and `configEnabledTOML`'s only caller is inside that same dead function. The remaining references were a test pinning the hand copy against itself. #170 had already moved every live decision to the catalog field, so the disagreement was latent and never reached a user. I asserted a user-visible consequence that did not exist, and a contributor could reasonably have acted on it.

The part I would have missed entirely is the Rust-side assertion:

```rust
assert_eq!(vendors[0]["enabled"], true);
```

The menu bar decides a vendor's state solely from that wire field, and `parseVendorCatalog` swallows a decode failure with `try?` and discards stderr. So a serde rename of `enabled` would empty the selector silently, passing every Swift test and every Rust test, because nothing asserted the *name*. Pinning both sides of a boundary that one side fails quietly across is the right instinct, and it is exactly the class of bug this project keeps producing — a frontend and Rust agreeing by convention rather than by anything that breaks.

Also good: deleting the test that pinned the copy against itself. A test asserting that a hand-maintained list matches a hand-maintained list is the kind that looks like coverage and provides none — it would have gone on passing while the catalog said something else.

Net −22 lines, zero dangling references (grep-verified across both Swift files, and the harness compiles them as one module so a miss would fail the build), full gate green.

PR #196, opened 2026-09-12
Let a [[custom]] provider borrow a built-in vendor's mark

GustavoBelo:
## Why

I watch two OpenCode Go accounts. `[opencode-go]` holds one key, so the second goes in as a `[[custom]]` entry against the same endpoint. That part works — the pointers into `usage.{rolling,weekly,monthly}` are stable and the rows pace fine. The mark doesn't. The native entry draws `opencode.svg`, the custom one falls back to its three-letter tag, and the same service ends up in the panel twice looking like two products.

`brandIconFile()` can't do better on its own. A custom entry arrives as `custom:<id>`, the id is free-form by design, and the switch has nothing to match on.

The field turned out to cover a second want. The mark doesn't have to be the endpoint's: my custom entry talks to OpenCode Go, but the account behind it runs DeepSeek models, and `brand = "deepseek"` is the mark I actually want on the bar. So this deliberately isn't "infer the provider from the URL" — the user names the mark, because only they know what the entry stands for.

## What this does

`brand = "<vendor slug>"` in a `[[custom]]` block names the built-in vendor whose mark the entry borrows — any of them, not only the one the endpoint belongs to. The report now carries `brand` on every entry — a built-in is its own brand — so the Omarchy model reads the provider off the report instead of mapping ids back to vendors, and still owns the artwork. Each frontend resolves the slug against whatever it ships; no file table moved into Rust.

`Config::validate` refuses a `brand` that isn't a built-in slug. A free-form value would reach the frontend as artwork it doesn't have and draw nothing, which is worse than the tag it replaced.

## Compatibility

- `brand` unset: unchanged.
- Old binary, new config: `CustomProviderConfig` doesn't deny unknown fields, so it's ignored. Verified against the released 1.16.0 binary.
- New binary, old frontend: one more JSON field.
- New plugin, old binary: no `brand` in the report, so `Model.js` falls back to the entry id and built-ins still get their marks.

## Tests

- `config.rs`: `brand` parses, defaults to `None`, and only a built-in slug is accepted.
- `report.rs`: a built-in relays its own slug, a custom entry relays the one it borrowed, and both reach the JSON.
- `model.test.mjs`: `brand` wins over the id, an uncovered brand degrades to the tag, and a report without `brand` still marks built-ins.
- Gate: `make test`, `cargo clippy --all-targets -- -D warnings`, `make changelog-check`, `cargo machete`, `omarchy plugin validate .`.


akitaonrails:
Thanks, Gustavo — the feature makes sense and the security model is sound: `brand` is limited to known built-in slugs and the frontend resolves it through a static asset table.

I updated the branch without rewriting your commits:

- merged current `main` to resolve the stale-base conflict;
- kept `TabSource::Custom` unchanged so this additive config feature does not break callers that construct or exhaustively match the public enum;
- omitted `brand` when unset, matching the documented tolerant `usage --json` contract;
- fixed Omarchy normalization, which had been dropping `brand` before QML could use it;
- added parser-path coverage and clarified that Omarchy is the currently supported branded frontend;
- formatted the Rust patch.

Local gates pass: changelog integrity, rustfmt, clippy with warnings denied, cargo-machete, Omarchy plugin validation, 1,574 Rust tests, and the GNOME/KDE/Windows/Omarchy model suites. I have approved CI for the adjusted head and will merge once the exact-head matrix is green.

akitaonrails (approved):
Approved after maintainer adjustments. Exact head 35c746785fffc330df414ba9a5c6e1c6c206b39a passed all eight CI jobs, and the local full/security gates are clean.

Issue #187, opened 2026-09-11
Feature request: Codex skill for checking AI usage


wellorbetter:
## Use case

While working in Codex, I'd like to invoke `$usage` and see quotas, balances, and reset times for all my configured AI providers directly in the conversation, without switching to another terminal or dashboard.

## Proposal

Provide a small Codex skill (`SKILL.md`) with installation instructions, reusing:

- `ai-usagebar usage --json` for usage, including `sections` to preserve balances and non-percentage rows.
- `ai-usagebar vendors --json` for configuration diagnostics.

Keep accounts, quota windows, and currencies separate, and show stale data and errors clearly. Missing dependencies or credentials should point to the existing setup instructions.

ai-usagebar would remain an external dependency and handle fetching, authentication, and caching. The skill would not read credential files, duplicate provider logic, or install anything during a usage query. No Rust API changes or native Codex `/usage` command are needed; `$usage` would invoke the installed skill.

## Scope

I read CONTRIBUTING.md and the thin-adapter guidance in CLAUDE.md, and couldn't find an existing issue or PR for this. I also saw the recommendation in #83 to keep additional integrations external.

Would you prefer a small skill and documentation in this repository, or a separately maintained skill linked under Community integrations?

akitaonrails:
Thank you for reading #83 before opening this — you found the precedent and asked the right question rather than assuming either answer.

**External repository, linked under [Community integrations](https://github.com/akitaonrails/ai-usagebar#community-integrations).** That is what #83 settled for the Noctalia widget, and `cosmic-applet-ai-usage` and `AI Usage for Noctalia` both live that way today. Consistency matters here, but so does the reason behind it: this project already keeps six frontends in sync — Waybar, TUI, GNOME, KDE, Omarchy and the macOS menu bar — and each one added is another surface that has to be re-checked whenever the report shape moves. A skill maintained by someone who uses Codex daily will stay better aligned with Codex than one I update when something breaks.

Your design is right, and worth saying so explicitly since it is the part people usually get wrong: reading `usage --json` and `vendors --json` without touching credential files or duplicating provider logic is exactly the boundary this project asks of a frontend. `sections` is the correct field for preserving balances and non-percentage rows — a skill that only read percentages would silently misreport every balance-only provider.

[...]

PR #197, opened 2026-09-13
docs: add Codex usage skill to community integrations

wellorbetter:
Adds [usage for Codex CLI](https://github.com/wellorbetter/ai-usagebar-codex-skill) to Community integrations, following the external-skill approach discussed in [#187](https://github.com/akitaonrails/ai-usagebar/issues/187#issuecomment-5639100744).

The independently maintained skill uses `ai-usagebar usage --json` and `ai-usagebar vendors --json` to show remaining quotas, balances, and reset times. It preserves ordered sections for balances and other non-percentage data. This PR adds only the community link and an Unreleased changelog entry.

Validation on Windows:
- `cargo test --all-targets --locked` (localhost bypasses the local proxy)
- `cargo fmt --all -- --check` and `cargo machete`
- GNOME, KDE, Windows popover, and Omarchy Node contract suites
- `scripts/check-changelog-immutable.sh` and `git diff --check`


akitaonrails:
Thanks for building this and coming back with the link — this is the outcome #187 was aiming for.

I read the skill before linking it, since the README will send people there. It keeps to the boundary it described: it runs only `ai-usagebar usage --json` and `vendors --json`, reads `schema_version`, treats unknown fields as additive, uses `sections` as authoritative (so balance-only providers are not reduced to missing percentages), treats every report string as data, and never touches credential files. That is the contract this project asks of a consumer.

One commit on top (`d551035`) removes the `CHANGELOG.md` entry. Community integrations are external projects linked from the README, and the earlier links — the COSMIC applet in `0445a8b`, the Noctalia widget — did not get one; the changelog records what ai-usagebar itself ships. It also matters mechanically here: an `### Added` entry is what the release process reads to choose a minor bump, so a README link would have decided the next version number.

Two notes for your repository, neither blocking the link:

- **It has no LICENSE file.** Public is not the same as reusable — without a license, nobody can legally fork or adapt it. If you want others to build on it, add one.
- **The skill does slightly more than "two query commands"**: the narrow proxy lookup (including the Windows Internet Settings keys) and the bundled Node renderer. Both are reasonable and neither reads secrets, but saying so in the skill's own README lets people decide before installing.

Merging once CI is green.

PR #198, opened 2026-09-14
fix: read Antigravity CLI OAuth session fallback

BergaBruh:
## What this changes

- Adds a read-only fallback to `~/.gemini/antigravity-cli/antigravity-oauth-token` when the Antigravity CLI session is unavailable through the OS keyring
- Reuses the existing credential parser, including the nested `token` shape
- Keeps the keyring as the preferred source and does not write credentials back
- Updates the Antigravity source label and related documentation

## Checklist

- [X] `CHANGELOG.md` entry under `## [Unreleased]` in the right category
- [x] Regression test fails without the file reader and passes with it
- [x] Docs updated where they mention what changed (`README.md`, `config.example.toml`, `docs/`)

## Testing

- Arch Linux x86_64, niri 26.04 (8ed0da4), noctalia v5.0.1
- CLI `agy` 1.2.2
- `cargo test --quiet` passed: 1575 library tests, binary/integration tests passed; 13 live tests were ignored as expected
- Targeted Antigravity source-label regression test passed
- `cargo fmt -- --check` passed
- `git diff --check` passed
- `cargo clippy --all-targets -- -D warnings` passed 
- `cargo machete` passed
- macOS and Windows runtime paths were not tested locally; they are covered only by compilation/CI


akitaonrails:
Audited — merging once CI is green. The design is right: the file is one an official CLI wrote, it is read only, the keyring stays first, and nothing is ever written back. That is the bar in `CONTRIBUTING.md`.

**Hermeticity, checked rather than assumed**, since `credential::read()` now reads a path under `$HOME` and the AUR `check()` runs `cargo test` during `makepkg`. I ran the suite with `HOME` pointing at an empty directory, and again at one containing a decoy `~/.gemini/antigravity-cli/antigravity-oauth-token`: 1593 passed both times and no test changed outcome. Every fallback test goes through `SavedCredential::Blob` / `Absent`; only the `#[ignore]`d live test reaches the real reader.

**One commit on top (`45a0570`): bound the file read.** `read_cli_token_file` did an unbounded `fs::read` and only then applied the 64 KiB cap, while the Credential Manager path checks the length before copying anything. A FIFO at that path was the sharper problem: opening one for reading blocks until a writer appears, and this runs inside the widget's fetch, which must always answer. It now checks `metadata(...).is_file()` *before* `open`, and reads through `take(MAX_BLOB_BYTES + 1)` so an oversized file still reaches `decode_blob_bytes` and is rejected there. Tests cover an oversized file, a directory, and a missing path. I also named the CLI token file in `CLAUDE.md`'s list of things Antigravity tests must never touch.

**A correction to the description, and a pre-existing bug it surfaced.** The PR says it "updates the Antigravity source label" and that a "source-label regression test passed", but `panels.rs` changes only a comment, and no label test was added — `Google API (app closed)` predates this PR.

That label has been wrong since v1.17.0: `fetch_snapshot_at` sends both `NoLocalServer` and `AgyMissingCsrf` into `fetch_remote`, which sets `AntigravitySource::Remote` unconditionally. So when `agy` is running but withholding its CSRF token — the exact case this PR is about — the TUI says the app is closed. This PR now documents that string in three places. Not asking you to fix it here, since it predates your change: right after merging I will relabel it `Google API` on main, including the doc lines you added.

akitaonrails:
Merged, and the follow-up I mentioned is in as `f1ffcce` on main: the source row now reads `Google API` rather than `Google API (app closed)`, which is true whichever reason the fallback fires for. That includes the doc lines this PR added, plus the stale "while nothing runs" doc on `AntigravitySource`. Thanks for the fix — the `agy` CSRF case now has a session to fall back on.

PR #200, opened 2026-09-16
fix(ollama): parse the monthly-quota response shape

mludoml:
## Problem

Some Ollama Cloud accounts (verified live, 2026-09-16) report:

```json
{
  "limits": {
    "monthly": {
      "usage": 0.003,
      "models": [{"name": "gpt-oss:120b", "request_count": 100}]
    }
  },
  "activity": {"cost": "0.00000", "period": {"type": "last_4_weeks"}}
}
```

instead of the documented `session` + `weekly` shape, under the same `"pro"` plan label (both shapes confirmed live against real Pro accounts). `Limits` only had `session`/`weekly` fields, so a monthly account parsed to `session: None, weekly: None` — no window to render. Every frontend (Waybar, TUI, Omarchy, GNOME, KDE) then fell through to a bare `"Ready"`/status pill with no percentage, cost, or model breakdown, even though the account has real usage data.

## Fix

- `ollama/types.rs`: `Limits.monthly: Option<Window>`, parsed into `OllamaSnapshot` the same way as `session`/`weekly`. Added `parses_live_captured_monthly_body`, a regression test using the real response shape above (fails on `main`, passes here).
- `usage.rs`: `OllamaSnapshot` gains `monthly: Option<UsageWindow>` and `monthly_models: Vec<OllamaModelUsage>`.
- `tui/panels.rs`: renders a "Monthly" metric section, its top-models block, folds it into `headline_pct` and `compact_cells`. The window has no fixed length or reset timestamp on the wire (same as `session`/`weekly` already), so it paces against a nominal 30 days — cosmetic only, there is no reset timestamp to pace against either way.
- `ollama/vendor.rs`: `severity()`, Waybar placeholders (`oll_monthly_pct`, `oll_monthly_reset`, `oll_monthly_pace`, `oll_monthly_pace_indicator`), and a tooltip row for the monthly window.
- Docs: `ollama-setup.md`, `vendor-endpoints.md`, `format-placeholders.md` now describe both response shapes.

Both shapes stay independent (never observed combined in one response), so this is additive — existing session/weekly accounts are unaffected.

## Testing

Ran the full CONTRIBUTING.md gate on this branch:

```
cargo test                                # 1577 passed, 0 failed
node gnome-extension/marker-logic.test.mjs  # passed
node kde-plasmoid/plasmoid-logic.test.mjs   # passed
node windows/popover/popover.test.mjs       # passed
node omarchy/model.test.mjs                 # passed
cargo clippy --all-targets -- -D warnings # clean
cargo fmt --all -- --check                # clean
cargo machete                             # no unused deps
```

Also manually verified end-to-end against my real Ollama Cloud Pro account (the monthly-only shape): built `--release`, ran `ai-usagebar usage --json`, confirmed the `ollama` entry now has a `Monthly` metric with `percent`, top models, and activity cost instead of an empty entry; confirmed in the Omarchy panel via screenshot that the bar/panel show `Monthly 0%` instead of `Ready`.

## Checklist

- [x] CHANGELOG.md entry under `[Unreleased]` / `Fixed`
- [x] Test that fails without the change (`parses_live_captured_monthly_body` against `main`'s `Limits`/`into_snapshot`)
- [x] No leftovers (existing session/weekly path untouched, mirrored the same pattern)
- [x] Docs updated (`ollama-setup.md`, `vendor-endpoints.md`, `format-placeholders.md`)


akitaonrails:
## PR #200 audit: parse the monthly-quota response shape

Trust gate: clear — no workflow, dependency, submodule, symlink, executable-bit or binary changes; no new network, credential or filesystem surface.
Head audited: `649dc7a`.
Local gates: `make test` (1579 passed, 0 failed, plus the desktop and Omarchy contract suites) and `cargo clippy --all-targets -- -D warnings` clean on my merged tree.

No findings. Recording what I checked, because two of these are the failure modes this project has actually shipped before:

- **The nominal 30-day window is not rendered as fact.** This was my main concern: a guessed window length presented as an exact countdown or pace is a bug we had to undo for Cursor in #162. It does not happen here. `Self::window` sets `resets_at: None`, and `pacing::calc` returns `Pacing::neutral()` at its first `let Some(reset) = reset else` before it ever reads the duration. So `oll_monthly_reset` stays empty and the pace glyphs stay neutral — the 30 days only ever describes the window's nominal length, exactly as the comment claims and exactly as `session`/`weekly` already behave. The comment in the diff is accurate, which is why I checked it rather than trusting it.
- **Stale caches still parse.** `OllamaSnapshot` derives no `Serialize`/`Deserialize`; the vendor cache stores the raw wire body and `parse_cache` re-parses it as `Body`, where `monthly` carries `#[serde(default)]`. A cache written by 1.17.1 therefore loads without error after upgrade instead of counting as an unparseable payload.
- **`severity()` folds the new window in** (`session.max(weekly).max(monthly)`), so a monthly account can actually reach the warning/critical classes rather than sitting permanently green.
- **Docs are complete.** `format-placeholders.md`, `vendor-endpoints.md` and `ollama-setup.md` are all updated; the five new `oll_monthly_*` placeholders are documented rather than left to be discovered from source.
- **Changelog** sits under `[Unreleased]` in the `Fixed` category, which is right for a shape the parser could not read.

### Claim ledger

|---|---|---|
| Some Pro accounts report `limits.monthly` instead of session/weekly | `LIVE_MONTHLY`, a real captured 200 body dated 2026-09-16, with `limits.monthly` present and both others absent | confirmed |
| The old parser rendered nothing for those accounts | `Self::window` returns `None` for absent windows, and the bar/tooltip have no window to show — reproduced by reading the pre-change projection | confirmed |
| The two shapes are mutually exclusive in practice | Stated as an observation, and the code does not depend on it — all three windows are independent `Option`s, so a response carrying both renders both | confirmed (and safely not relied upon) |
| Tests cover the new shape | `parses_live_captured_monthly_body` asserts the monthly window, the model breakdown and the activity cost; the existing session/weekly test still asserts `monthly.is_none()` | confirmed |

### Pros

- Fixes a silent, total failure: affected accounts saw a bare "Ready" pill with no percentage anywhere.
- Built on a real captured body rather than a guessed schema, and the capture is committed as the fixture.
- Respects the project's rule about not inventing a denominator or a reset time — the one place it would have been easy to fabricate a number, it declines to.
- Docs and changelog landed with the code, not after.

### Cons

- The monthly window carries no reset, so its row shows a percentage without a countdown. That is honest rather than ideal; the JSON route simply does not carry the date. If the HTML UI's reset date is ever worth having, it is a separate change.

Recommended action: **merge as-is.** I will refresh the branch against `main` first — it branched before v1.17.1, so its `[Unreleased]` still lists two entries that have since shipped, and merging without resolving that would re-list released entries.

akitaonrails:
Refreshed against `main` in `a6829c4`. Your code is untouched; one maintainer commit was needed, and it is worth describing because the merge looked perfectly healthy.

Merging `main` into this branch produced **no conflicts at all**. That is precisely the failure mode this project has been bitten by four times now: the branch predates v1.17.1, so your bullet sat under `[Unreleased]`, and git merged it cleanly into whatever now occupies that position — which is the **published `## [1.17.1]` section**. The result was a shipped release silently gaining a feature it never contained, with no conflict marker and no removed line for a `grep '^-'` style check to notice.

`make changelog-check` caught it by diffing the section against tag `v1.17.1` byte for byte:

```
error: the published [1.17.1] section no longer matches v1.17.1:
    15a16,24
    > - **Ollama Cloud monthly-quota accounts no longer show a bare "Ready".**
    ...
```

Moved back under `[Unreleased]` in `a6829c4`. Nothing about your change was wrong — this is a property of how git merges append-only lists, and it is why a clean merge is not treated as evidence here.

Gate on the merged tree: `make test` (**1580** passed, 0 failed, plus the GNOME/KDE/Omarchy/popover contract suites), `cargo clippy --all-targets -- -D warnings` clean, `cargo machete` clean, `make changelog-check` clean. Version files verified still at 1.17.1 across `Cargo.toml`, `manifest.json` and both PKGBUILDs — the other shape of this same hazard, where a stale branch quietly reverts them.

Merging once hosted CI is green. Thanks for the captured response body; committing the real payload as the fixture is what made this reviewable rather than speculative.

PR #201, opened 2026-09-16
fix(macos): distinguish disabled providers and offer explicit enable

samirhvbr:
A provider with credentials but `enabled = false` appears as “Configured” in macOS Preferences while disappearing from the provider selector. For Claude, the offered action is to sign in again, which does not address the disabled setting.

Preferences now show “Disabled” (with a credential-available hint when applicable) and an explicit Enable button. Once enabled, the row resumes its usual credential status and sign-in action. Enabling does not change the primary provider or switch accounts.

The button calls the new `ai-usagebar settings enable <vendor>` command. Rust uses the existing comment-preserving, atomic config writer and changes only the selected provider's enabled flag. Unlike automatic discovery, this explicit opt-in can override `enabled = false`. Repeated calls do not rewrite an unchanged file. The UI disables buttons while working and reports failed launches or nonzero exits; older binaries fail visibly with an update hint. No credentials are passed through the subprocess.

Validation on macOS:

- `make test`: Rust tests and GNOME, KDE, Windows popover, and Omarchy contracts passed.
- `cargo clippy --all-targets -- -D warnings`, `cargo fmt --all -- --check`, and `cargo machete` passed.
- `bash macos/run-tests.sh` passed, including the enabled/configured combinations and subprocess argument, failure, and missing-executable cases.
- The new preference assertions fail four times against the upstream status/button logic (extracted unchanged into testable helpers), and pass with the fix.
- Built the CLI and exercised `--config` with a temporary path containing spaces: exact preservation of other settings/comments, idempotent writes, and rejection of malformed TOML passed.

The native code was compiled and exercised through the Swift harness; no manual GUI click-through is claimed. Documentation and an Unreleased changelog entry are included.


akitaonrails:
## PR #201 audit: distinguish disabled providers and offer explicit enable

Trust gate: clear — no workflow, dependency, submodule, symlink, executable-bit or binary changes; no new network or credential surface.
Head audited: `f879fc3`, plus maintainer commits `63eee10` and the `main` merge, now `885cc5c`.
Local gates on the merged head: `make test` (1579 passed, 0 failed, plus the GNOME/KDE/Omarchy/popover contract suites), `cargo clippy --all-targets -- -D warnings` clean, `make changelog-check` clean.

### Findings

- `[SHOULD-FIX] macos/ai-usagebar-menubar.swift:1461,1513` — **fixed in `63eee10`.** The button read `Re-logar` and the spinner `verificando…`. To be clear about provenance: this PR did not introduce them, it carried them forward from the code it replaces — they have been on `main` since well before this branch. The UI is English everywhere else, so they now read `Sign in again` and `checking…`, and the assertion at `ai-usagebar-tests.swift:1051` moved with them. The same three strings were live in the GNOME preferences; those are fixed separately on `main` in `1ff11fc`, since this PR does not touch that file.

- `[NIT] macos/ai-usagebar-menubar.swift:1466-1470` — `vendorButtonLabel` grows a small per-vendor arm list (`antigravity`, `cursor`, `kiro`). Rust already owns this knowledge in `VendorId::sign_in_hint()`, but the vendor catalog JSON does not currently expose it, so wiring it through is a larger change than this PR. Not blocking, and the existing `pkg` switch above it sets the precedent. Worth a follow-up that adds a hint field to the catalog and deletes both tables.

No other findings.

### Claim ledger

|---|---|---|
| Disabled providers are distinguishable from unconfigured ones | `vendorStatusText` returns `Disabled` / `Disabled — credential available` before any credential branch; asserted both ways in the new test | confirmed |
| `settings enable` preserves other settings | `explicit_enable_preserves_other_settings_and_is_idempotent` writes a config with a comment, an unrelated `api_key` and a second `enabled = false`, and asserts byte-equality against a single targeted replacement | confirmed |
| Enabling is idempotent | same test calls it twice and compares; the implementation also skips the write when the document is unchanged | confirmed |
| Does not overrule an explicit opt-out by accident | `enable_vendor_at` is reached only from the explicit `settings enable` path, and the doc comment marks the distinction from discovery, which still respects an existing `false` | confirmed |
| Write failures stay visible | `enableVendor` returns a message on spawn failure and the view renders it in red | confirmed |

### Pros

- The failure it fixes is real and silent: a disabled provider and a provider missing a key looked identical, so the honest remedy was invisible.
- The config write is `toml_edit`-based and surgical — comments, formatting and unrelated keys survive, which is the part most likely to have been done destructively.
- Tests cover the malformed-config and missing-config paths, not just the happy one.
- Docs updated alongside (`docs/configuration.md`, `macos/README.md`).

### Cons

- The Swift per-vendor arms duplicate knowledge Rust already holds (nit above).
- The macOS test suite only runs on macOS CI, so my local gate proves the Rust half; the Swift half rests on the hosted run.

Recommended action: **merge**, after the maintainer commits above.

PR #202, opened 2026-09-16
fix(macos): show named Codex accounts in provider selectors

samirhvbr:
The Rust backend supports `[[openai.accounts]]`, but the macOS provider list expands only Claude and OpenRouter accounts. A named Codex login therefore cannot be selected in the menu, Preferences, Overview, or vendor-cycle shortcut.

Read Codex labels through the existing account-label parser and add `openai@<label>` entries to the shared provider list and Preferences. Display names use the Rust catalog. The existing argument mapper sends `--vendor openai --account <label>`; Rust continues to resolve auth paths and isolate account caches. Named accounts remain available without a default Codex login. Disabled Codex still hides every account, and selecting usage does not switch the active Codex login.

Validation on macOS:

- `bash macos/run-tests.sh` passed: label parsing/order, display names, missing default login, disabled provider, Overview filtering, shortcut cycling, Preferences, and subprocess arguments.
- Removing the Codex expansion from the selector and Preferences while retaining the test seams causes four assertions to fail; they pass with the change.
- `make test`, `cargo clippy --all-targets -- -D warnings`, `cargo fmt --all -- --check`, and `cargo machete` passed.

Documentation and an Unreleased changelog entry are included. Native code was compiled and exercised through the Swift harness; no live OAuth requests or manual GUI click-through were performed. This branches directly from upstream main and does not depend on #201.


akitaonrails:
## PR #202 audit: include named Codex accounts across provider selectors

Trust gate: clear — no workflow, dependency, submodule, symlink, executable-bit or binary changes. Swift, docs and changelog only.
Head audited: `ad7e4d2`. Hosted CI green on that head.

No findings. What I verified:

- **No credential exposure.** `codexAccountLabels()` reads `config.toml` and passes it straight to `accountLabels(inTOML:vendor:)`, which returns label strings only. Nothing from that file is logged, rendered or put into a process argument — the same shape `claudeAccountLabels()` already uses on `main`, so this adds no new handling of a file that may contain `api_key` values.
- **It follows the established precedent rather than inventing one.** Named-account support for Anthropic and OpenRouter already worked this way; Codex was the odd one out. `preferenceVendorIds` now routes all three through one function instead of three parallel paths.
- **Rust stays the source of truth for display names.** `entryDisplayName` takes `Codex · work` from the catalog, so the frontend is not carrying a provider-name table.
- **`show_default_account` semantics are preserved**, with the Rust behaviour spelled out in a comment at the call site and asserted in the tests.

### Claim ledger

|---|---|---|
| Named Codex accounts were missing from the selectors | `preferenceVendorIds`' switch had no `openai` arm before this change | confirmed |
| Labels preserve config order and exclude other providers | `testCodexAccounts` asserts `["work", "personal"]` exactly | confirmed |
| The keyboard shortcut cycles into and out of named accounts | `nextVendorId` asserted in both directions, including the wrap | confirmed |
| Vendor args map correctly | `vendorArgs(for: "openai@work")` asserts `["--vendor", "openai", "--account", "work"]` | confirmed |
| Disabled providers stay out | asserted via the `disabled` catalog case | confirmed |

### Pros

- Removes a real inconsistency: a user with `[[openai.accounts]]` configured could see the accounts in the TUI but not select them in the menu bar.
- Test coverage is unusually complete for a UI change — ordering, filtering, cycling, wrapping, display names and argument mapping are all pinned.
- `macos/README.md` updated alongside.

### Cons

- The macOS suite runs only on macOS CI, so the hosted run is the sole executor of these assertions; my local gate does not cover Swift.

Recommended action: **merge.** I will refresh against `main` first — this branch predates v1.17.1, so its `[Unreleased]` still carries entries that have since shipped, and it touches the same Swift files as #201, so it is merged second to resolve once.

akitaonrails:
Refreshed against `main` in `17fc90f` — no changes to your code, but two conflicts needed resolving and both are worth describing since they were not trivial:

1. **`CHANGELOG.md`** — this branch predates v1.17.1, so its `[Unreleased]` section sat exactly where the released 1.17.1 section now lives. Git's conflict region included the `## [1.17.1] — 2026-09-16` heading itself, so a careless resolution would have folded two already-released entries back into `[Unreleased]`. Resolved keeping your bullet plus the two from `main`, with the released section untouched; `make changelog-check` confirms every published section still matches its tag byte for byte.

2. **`macos/ai-usagebar-tests.swift`** — your `testCodexAccounts()` and #201's `testEnableVendorCommand()`/`testDisabledVendorPreferences()` were appended at the same spot, and both sides relied on the same trailing brace, so a naive union would have dropped one. Resolved by giving each block its own closing brace and verifying mechanically: the file's brace balance is 0, and all 20 `func test…` definitions have exactly 20 matching invocations in `TestRunner`.

One incidental change reaches your diff from `main`: the button label your test asserts is now `Sign in again` rather than `Re-logar` — Portuguese that had been on `main` since before either of your branches, fixed in #201.

Local gate on the merged tree: `make test` (1579 Rust tests plus the GNOME/KDE/Omarchy/popover suites), `cargo clippy --all-targets -- -D warnings`, and `make changelog-check` all clean. Merging once hosted CI is green — the macOS job is the one that actually executes your Swift assertions.

Issue #203, opened 2026-09-16
Antigravity: local 401 "missing CSRF token" (newer agy CsrfInterceptor) is reported as signed out and hides the working Cloud Code fallback


rendermani:
## What happens

ai-usagebar (1.14.0) shows `HTTP 401` for Antigravity instead of quota numbers when a newer `agy` build is running locally.

## Measured root cause

The local `agy` language server (measured: pid 5271, ports 43865 RPC / 38881 TLS) answers `RetrieveUserQuotaSummary` — and `GetUserStatus` — with:

```
HTTP 401
{"code":"unauthenticated","message":"missing CSRF token"}
```

Newer `agy` builds wrap the JSON-RPC surface in a `CsrfInterceptor`. ai-usagebar scrapes the CSRF token out of the HTML at `/` (`fetch_csrf`, splitting on `csrfToken":"`); these builds no longer expose one there, so the RPC goes out without the `x-codeium-csrf-token` header and is rejected.

In `src/antigravity/fetch.rs`, `is_actionable()` classifies any 401/403 as "server is up but signed out — the user can act on it". That 401 therefore wins `select_probe_error()`, `fetch_snapshot_at` keeps the diagnosis local, and the Cloud-Code fallback is never tried.

The fallback itself works correctly when it is reached: a refreshed keyring session (`gemini`/`antigravity` in the Secret Service) against `daily-cloudcode-pa.googleapis.com/v1internal:retrieveUserQuotaSummary` returns real quota buckets (HTTP 200) with User-Agent `antigravity`.

## The problem in one sentence

A 401 whose body says the *probe* lacked a CSRF token is not "signed out" — the server is up and signed in — so it must fall through to the cloud path instead of being reported as an auth failure.

## Fix

On branch `fix/antigravity-csrf-fallback` (local; no fork, hence this issue instead of a PR):

- `is_csrf_rejection()`: a 401 whose body contains `csrf token` (case-insensitive) is a rejection of the probe, not the session. Matched on the body, not the status, so a genuine signed-out 401 keeps its own diagnosis.
- `is_actionable()` excludes that case, so it no longer ranks as "signed out".
- `select_probe_error()` gives it its own rank: below a genuine 401/403 (a real auth rejection still outranks it), above transport noise and the TLS echo — `fetch_snapshot_at` routes to the remote path based on the *selected* error, so a connection refusal standing in its place would silently cost the fallback routing.
[...]

Issue #195, opened 2026-09-12
antigravity: local probe fails with HTTP 401 ("missing CSRF token") when agy CLI is running, blocking remote fallback


Josemalyson:
### Summary
When running an interactive `agy` CLI session (or multiple `agy` sessions), `ai-usagebar --vendor antigravity` fails with:
`HTTP 401: authentication rejected — credentials may be missing, expired, or invalid`
even though the saved Google session in the OS keyring (`secret-tool`) is 100% valid and active.

### Root Cause
1. `is_antigravity_process()` in `src/antigravity/fetch.rs` matches `comm == "agy"`.
2. `discover_ls_ports()` finds the ports opened by `agy` on loopback (e.g., ports 42423, 35065).
3. `fetch_csrf()` attempts to retrieve the CSRF token via `GET http://127.0.0.1:<port>`. However, unlike the Antigravity IDE GUI webview, `agy`'s embedded HTTP server returns `404 Not Found` for `GET /`, so `fetch_csrf()` yields `None`.
4. `post_rpc()` sends `GetUserStatus` without the `x-codeium-csrf-token` header.
5. `agy` responds with:
   ```json
   {"code": "unauthenticated", "message": "missing CSRF token"}
   ```
   with HTTP status `401`.
6. `select_probe_error()` treats status `401` as actionable (`is_actionable()` returns true), so `open_session()` returns `Err(AppError::Http { status: 401, body: "{\"code\":\"unauthenticated\",\"message\":\"missing CSRF token\"}" })`.
7. Because `is_no_local_server(&e)` only matches `NO_LOCAL_SERVER`, `fetch_snapshot` does NOT fall back to `Origin::Remote` (reading the keyring and querying Google Cloud Code API).
8. The 401 error is persisted in `.last_error` and displayed to the user as an authentication rejection.

### Expected Behavior
If a local server probe fails authentication specifically due to missing CSRF token (e.g. `body.contains("missing CSRF token")`), it should allow fallback to `Origin::Remote` (the saved session in the OS keyring) rather than halting with a false-positive credential rejection.

### Environment
- OS: Linux (Fedora 43 / GNOME 50.4 on Wayland)
- Binary: `ai-usagebar v1.16.0`
- Process: `agy` CLI

PR #204, opened 2026-09-16
Antigravity: fall through to the cloud path on a CSRF-missing 401

rendermani:
## Summary
- The local agy language server's RetrieveUserQuotaSummary now carries a CsrfInterceptor on newer builds, and rejects the probe with `401 {"code":"unauthenticated","message":"missing CSRF token"}`.
- `is_actionable()` in `src/antigravity/fetch.rs` ranked that 401 as "signed out, user can act on it", so it won `select_probe_error()` and the working Cloud-Code fallback (`daily-cloudcode-pa.googleapis.com/v1internal:retrieveUserQuotaSummary`, User-Agent `antigravity`, via the OS-keyring Google session) was never tried.
- This fixes the classification: a CSRF-missing 401 from the local probe now falls through to the cloud path instead of being reported as signed-out.
- Added a test that pins the exact CSRF-missing-token response body.

## Test plan
- New unit test pinning the CSRF-401 body and asserting it falls through rather than winning `select_probe_error()`.
- Built with the toolchain at `~/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/bin`.

---

## Second commit: keyring read behind the first fix (8910b32)

The CSRF fix alone was not enough on the reporting machine: the cloud path it now falls through to failed on the **next** step. Measured:

```
$ secret-tool lookup service gemini username antigravity
secret-tool: secret does not contain a textual password   # exit 1
$ secret-tool search --all service gemini username antigravity
[/3]
label = Password for 'antigravity' on 'gemini'
secret = {"token":{...}}                                   # full session JSON
```

`agy`-built Antigravity stores the keyring item with a **non-textual content type**, so `lookup` refuses it while `search` hands it over. `credential::read()` used `lookup`, concluded "no saved session", and reported *"no local server found … or sign in to Antigravity once"* — both halves wrong: the local server was up (127.0.0.1:43865/38881) and the session existed and was valid.

Fix: `read_platform` (Linux) falls back to `secret-tool search --all` when `lookup` fails, parses the `secret = ` line (pure parser, pinned by a test on the verbatim block shape), and feeds it through the existing decoder/parser. `lookup` is still tried first, so textual items behave exactly as before.

Verified with real numbers after both fixes — `ai-usagebar usage --json`, antigravity entry, source `Google API (app closed)`:

- Gemini 5h: 0% (resets in 3h 59m)
- Claude & GPT OSS 5h: 0% (resets in 4h 59m)
- Gemini weekly: 0% (resets in 6d 22h)
- Claude & GPT OSS weekly: **100%** (resets in 1d 16h)

### Reproducibility note for the maintainer

This exact failure chain (local 401 hiding the cloud path) was found and is reproducible on a machine where `agy` runs as a **background service** (`agy-quota-keeper.service`, a systemd user unit keeping an idle `agy` session alive specifically to serve the quota RPC for ai-usagebar). Without such a service — i.e. on a machine where Antigravity only runs as an interactive app — the local probe usually succeeds and neither bug is visible. To reproduce: run an `agy` language server in the background whose build carries the `CsrfInterceptor`, and have the keyring item stored the way `agy` stores it (non-textual content type).


akitaonrails:
Thank you for writing this up with the measured body and the ranking reasoning — it is a correct fix. It is also already on `main`: **#195 shipped the same change in v1.17.0**, and this branch is based on `9bd5e60`, which predates it.

Concretely, `main` already has `is_missing_csrf()`, its exclusion from `is_actionable()`, the dedicated rank in `select_probe_error()` (`actionable.or(missing_csrf).or(last).or(echo)` — below a genuine 401/403, above transport noise and the TLS echo), the `Origin::Remote` routing, and four tests including the signed-out counter-case. I checked each of those against your description rather than assuming overlap, because your ranking argument is the part that is easy to get wrong and I wanted to be sure it was covered and not merely adjacent.

The one place the two differ: `main` matches the body **structurally and exactly** (`code` = `unauthenticated` and `message` = `missing CSRF token`) where this branch matches a case-insensitive `csrf token` substring. The stricter form is deliberate — an unrelated local service answering 401, or a genuinely signed-out Antigravity product, must not become permission to send the saved Google session to Google. Your measured body satisfies the strict match, so nothing is lost for the case you hit.

Closing as superseded rather than merged, since rebasing 254 lines onto code that already does this would be work for no change in behaviour. If you try v1.17.1 and it still reports `HTTP 401`, that means an `agy` build whose body differs from the one you measured — post that exact body in #203 and it is a one-line widening of the matcher, with your existing reasoning intact.

I have closed #203 with the same explanation.

PR #205, opened 2026-09-16
feat(claude-desktop): account merge-history for side-by-side Desktop profiles

ohmaseclaro:
## What this changes

Adds `account merge-history`, which merges every account's sessions and
schedules into whichever account a given Claude Desktop profile is signed
into:

    ai-usagebar account merge-history --data-dir <DIR> [--from <DIR>]... [--dry-run]

Claude Desktop can be launched against an alternative profile with
`--user-data-dir`, which is how more than one account can run at once as
separate apps. Each such profile has its own history, so those windows do not
see each other's conversations. That is the gap this fills.

`account switch` cannot serve this case, for two reasons that are structural
rather than cosmetic:

- `plan_switch` requires a saved profile per label and installs that label's
  stored credential into the profile, overwriting the live login of a profile
  that is already signed in;
- `apply_switch` begins with `app.quit()` and ends with `app.relaunch()`,
  targeting the app by *application name*. Once several app bundles share a
  `CFBundleName`, that is ambiguous — it may quit a different profile's app, or
  the default one.

`merge-history` swaps no credential and never controls the app.

### Why `--data-dir` has to reach `Paths`

`Paths::at` already takes every root explicitly, and everything downstream
(`sessions_root`, `config_json`, `synced_path`) derives from those fields — so
the merge itself was already directory-agnostic. Only `Paths::resolve`
hardcodes the profile location. The new `Paths::for_data_dir` is that
constructor with the profile supplied by the caller. `Paths::resolve` and every
existing call site are untouched, so behaviour without the new subcommand is
byte-identical.

`merge.rs` is unchanged apart from widening one `const` to `pub(super)`. The
command is assembled from the existing public functions — `plan_session_merge`,
`plan_scheduled_merge`, `plan_name_convergence`, `current_state`,
`canonical_routines` — so there is still exactly one implementation of the
merge and no duplicated semantics.

### Why `backups_dir` is a per-profile sibling

This is the part that is hard to re-derive, so it is worth stating. With
`--data-dir D`, `profiles_dir` stays shared (it is the label-keyed credential
store, and the same accounts back every profile) but `backups_dir` becomes
`D-ai-usagebar/backups` — a sibling of the profile, never inside it.

`synced_path()` derives from `backups_dir`, and that ledger records what each
account held *after the last merge*; `deletion_candidates` is the difference
between live state and it. If several live profiles shared one ledger, each
merge would overwrite it with that profile's own, narrower view: profile A
records account X with ten routines, profile B — which holds only six of them
locally — then records six, and A's next run sees four as intentionally
deleted. Those phantom candidates are kept automatically in a non-interactive
run, but they would surface in an *interactive* `account switch`, where they
can be answered and therefore deleted for real. Hence one ledger per profile.

### Safety properties

- **Additive.** No deletion sweep runs, and `--delete-conflict` has no effect
  here. Conflicting items are reported and kept. (The existing
  `resolve_deletions` already keeps everything when stdin is not a terminal,
  so an unattended run could not delete anyway; this command does not offer
  the path at all.)
- **Sources are read-only.** Only `read`/`read_dir`/`metadata` touch them,
  every write destination is checked to lie under the target's
  `sessions_root`, and sources are not required to be idle — reading an index
  out of a running profile is safe, writing into one is not.
- **Never two writers.** The target profile must not be running. Detected from
  the process's own `--user-data-dir` argument, never the app name or bundle
  id, and it fails closed if `ps` cannot be read.
- **The default profile is refused as a target**, comparing symlink-resolved
  paths so a trailing slash or `..` cannot slip past. It remains usable as a
  read-only source.
- **Idempotent.** An index is copied only when the destination is absent or
  strictly older, so re-running is a no-op. The target's own
  `scheduled-tasks.json` is placed only when absent, so staging cannot drop a
  routine that exists only there before the merge reconciles it.
- **Partial failure is loud.** A source that exists but cannot be read is
  listed and the command exits non-zero: an absent source and an unreadable
  one must not look alike.

### One behaviour worth calling out

It deliberately crosses accounts. After a merge, the window signed into one
account lists conversations and routines started under the others — that is
the point, but it means one account's window shows another's chat titles. Said
plainly in `--help`, in `docs/claude-accounts.md` and in the changelog, so it
is not later read as a leak.

## Checklist

- [x] `CHANGELOG.md` entry under `## [Unreleased]` → `### Added`
- [x] Tests: three for the liveness predicate, including a profile path
      containing a space. The first version split the `ps` line on whitespace,
      which truncated the argument and reported a *running* profile as idle —
      a guard failing in the direction that permits the corruption it exists
      to prevent.
- [x] Docs updated — `docs/claude-accounts.md` gains a "Side-by-side profiles"
      section and a row in the setup table
- [x] No helper left without callers

## Testing

macOS 15 (darwin 25.5.0), Apple silicon, Claude Desktop 1.52386.6.

`cargo test` (1607 passing), `make test`, `cargo clippy --all-targets -- -D
warnings`, `cargo fmt --all -- --check`.

Exercised against real profiles: a relocated profile with no history went from
0 to 509 session indexes and 1 to 4 account directories in one run; the second
and third runs staged nothing and reported "already up to date". After a full
merge, zero files in the default profile were modified and no ledger, lock or
sibling directory was created beside it. Both guards were confirmed to fire —
the default profile refused as a target, and a target whose app was running
refused — as was the non-interactive keep-everything path.

Linux and Windows are unaffected: the subcommand is macOS-only in the same way
the rest of the Desktop support is, and returns the existing "no Claude
Desktop app data" error elsewhere.


ohmaseclaro:
Notes in progress, while this is still a draft.

Added a fourth commit: default `--from` sources now also have to hold at least
one `<account>/<org>` directory, not merely a `config.json`. A profile the app
has opened once already has the config, so abandoned and never-signed-in
profiles were being listed as sources and reported on while contributing
nothing. Explicitly named `--from` sources are deliberately still unfiltered —
asking for a source and being told it is empty is information.

PR #218, opened 2026-09-19
fix(usage): exit 0 after a complete error document (#217)

### LINKED ISSUE #217: `usage --json` exits 1 while printing a complete document, so the exit code can't be used by scripts
## Summary

When every configured entry fails, `ai-usagebar usage --json` prints a complete, well-formed document (each entry carrying its own `error`) and then exits 1. So the exit code and the payload disagree: the command did what it was asked to do, and the errors it's reporting are *data*, not a failure of the command.

That bit me in a small status script. I had it written as:

```sh
json="$(ai-usagebar usage --json)" || die "usage --json failed"
```

which is the obvious way to write it, and it went silent exactly when it mattered most — the machine where all four accounts were broken at once was the machine where I most wanted the per-account diagnosis. I now ignore the exit code and gate on `has("entries")` instead, which works but feels like working around a contract rather than using one.

## Reproduce

Any config where every entry errors (here: the Anthropic credential files were unreadable by the calling process):

```
$ ai-usagebar usage --json | jq '.entries | length'
8
$ ai-usagebar usage --json >/dev/null; echo $?
1
```

## What I'd suggest

Exit 0 whenever the document is complete, since per-entry failures are already represented inside it, and reserve a non-zero exit for "I could not produce the document" (bad config, unparseable TOML, etc.).

If the current behavior is deliberate — "non-zero means at least one vendor is unhealthy", which is a reasonable contract for a shell-driven status bar — then documenting it in `--help` would be enough for me, and I'll close this. Happy to send a PR for whichever you prefer.

## Environment

- ai-usagebar 1.10.0 on macOS 27.0 (also present in the 1.20.1 source)

viniciusdsandrade:
## What this changes

Fixes #217.

`ai-usagebar usage` (plain or `--json`) now exits 0 after printing a complete document, even when every entry carries its own `error`. Per-entry fetch/auth failures stay data inside the document. Non-zero remains only when the command cannot produce the document (missing or unreadable `--config`, unparseable TOML, no vendors enabled, or a runtime/bootstrap failure).

This is a **deliberate contract change**, not an accidental bug. Current `main` and its test encode “exit 1 only when every entry failed,” and the binary comments that `usage` “deserves a real exit code.” Scripts that treated all-error as command failure will now see 0; they should inspect `.entries[].error` instead. The issue author preferred this contract (complete document → 0). Review can reject it if the maintainer prefers documenting the old policy instead.

The replacement test `produced_document_exits_zero_even_when_every_entry_failed` **fails on current `main`** (`left: 1`, `right: 0`) and passes on this branch. `--help` and the README scripting notes state the new exit contract. `collect_json` / in-process frontends are unchanged (they return `Result`, not a process status). The Waybar widget’s always-exit-0 contract is untouched.

## Checklist

## Testing

Ran on **macOS 26.7** (Darwin 25.6.0 arm64):

- `cargo test --lib produced_document_exits_zero_even_when_every_entry_failed` — fails on `main` (all-error document currently exits 1); passes here (all-error, mixed, and success documents exit 0; empty is not claimed as success)
- `cargo test --lib usage_help_states_complete_document_exits_zero`
- `make test`

Real binary, scratch `--config` (no user HOME config or credentials):

- `ai-usagebar usage --json --config` a TOML with default-on vendors disabled and one `[[custom]]` provider with no key — launched **twice**; both times exit **0**, stdout is parseable JSON with a non-empty `entries` array, each entry has a populated `error`
- Same binary against every built-in vendor `enabled = false` and no custom providers — exit **1**, stderr `no vendors enabled`
- `ai-usagebar usage --help` states that a complete document exits 0


viniciusdsandrade:
Pushed two revision commits.

**`fix(ci)` — why a CI fix is in this PR:** the ubuntu job ("Published changelog sections and versions are intact" = `make changelog-check`) has failed on **main itself** since the v1.20.1 commit, so this PR could not go green regardless of its own changes. What happened: v1.20.1 folded the never-published v1.20.0's entries into its own section (its Fixed entry records "that tag remains unused"), which the immutability guard read as the merge-deletion hazard it exists to catch. Restoring the `[1.20.0]` section would duplicate every entry under two versions and trip the `make test` duplicate-entry guard, so the script now carries an explicit allowlist of never-shipped tags. The same run also fixed its version-file extraction, which used a GNU-only `sed` form that matched nothing on BSD sed — on macOS every version file reported as unreadable. Both hazards the guard exists for were re-verified as still caught (tampered published section → error; deleted released heading → error) before pushing.

**`docs(usage)`** — moved the changelog entry from Fixed to **Changed** (it is a deliberate contract change; repo convention puts behavior changes there) with a migration note, and added the canonical health gate to the README: `ai-usagebar usage --json | jq -e 'all(.entries[]; .error == null)'`.

Local gate on macOS: `make changelog-check` (now actually runnable on macOS and green), `make test`, `cargo fmt --all -- --check`.

PR #225, opened 2026-09-20
fix(antigravity): abort candidate probing immediately on missing CSRF token

lhzapata:
## What this changes

When Antigravity candidate port discovery finds a local language server running (such as `agy`), probing candidate ports with plain HTTP can encounter a 401 unauthenticated response with `missing CSRF token`.

Because `agy` does not publish a token-discovery endpoint and requires an undiscoverable CSRF token, probing subsequent candidate ports (particularly companion TLS listeners on adjacent ports like `port + 1` from the same process or discovery group) causes plain HTTP requests to hit TLS endpoints. This triggers spurious and noisy host errors:
```
http: TLS handshake error from 127.0.0.1:...: remote error: tls: unrecognized name
```

This change aborts the candidate probing loop in `open_session` immediately upon encountering `is_missing_csrf`:
- The active language server requiring CSRF has already been identified; subsequent candidate ports belong to the same session or companion listeners and cannot provide unauthenticated metrics.
- Sparing the remaining candidate ports eliminates the spurious TLS handshake error logs on the host.
- The actionable missing-CSRF error is retained by `select_probe_error`, allowing `run` to fall back cleanly to the persisted Google OAuth session as intended.

## Checklist

## Testing

Tested on **Windows 11** (x86_64-pc-windows-msvc):

- Added unit regression test `agys_missing_csrf_aborts_further_candidate_probes_to_spare_tls_listeners` in `src/antigravity/fetch.rs`. With unpatched probing logic, mock `server2` is probed (`expect(0)` fails with expectation mismatch). With this patch, `server2` is never probed and the test passes (`expect(0)` asserted).
- `cargo test antigravity`: all 125 tests passed.
- `cargo test changelog`: guard tests passed (`no_changelog_entry_appears_under_two_versions`, `no_changelog_section_repeats_a_category`).
- `node gnome-extension/marker-logic.test.mjs`, `node kde-plasmoid/plasmoid-logic.test.mjs`, `node windows/popover/popover.test.mjs`, `node omarchy/model.test.mjs`: all passed.
- `cargo fmt --all -- --check`: passed cleanly.
- Re-tested with live running `agy` process: verified that spurious `http: TLS handshake error` console logs cease and Google OAuth session fallback resolves cleanly.


lhzapata:
Dealing with Antigravity agy harness particularities is so tiresome.

lhzapata:
Follow-up to my own patch: reviewing the multi-product probe ordering, I realized the original `break` was broader than intended. `probe_order` emits rank-0 ports from every discovered pid before rank-1 ports and then flattens the pid away, so the candidate right after an `agy` missing-CSRF answer can be another product's valid RPC listener — not agy's companion TLS port. A global break there would skip a signed-in IDE/desktop server entirely.

The new commit keeps the same goal — spare the TLS companion that generates the `http: TLS handshake error` noise — but scopes it per process: candidates now carry their owning pid, and a missing-CSRF verdict blocks only that pid's remaining listeners. Other products keep being probed.

Coverage: a same-pid regression test (second listener never probed), a cross-pid regression test (a valid server after missing CSRF is used), and a live Windows check with `agy` plus two IDE `language_server` processes in parallel — all three pids answered missing CSRF, all three TLS companions were skipped, probing continued across processes, and the OAuth fallback returned real quota with zero new handshake errors in the `agy` console.

